PR #71, opened 2018-01-16
docs: how to build a plugin

dan-ziv:

How to build a plugin. 

### CheckLists

- [x] Docs have been updated


dan-ziv:
@AnnaAle

Thanks,
PR is fixed recording to your comments.

docs/how-to-build-a-plugin.md
```diff
@@ -0,0 +1,475 @@
...
+</script>
+</body>
+</html>
+```
+
+**That's it! the player starts to play and your plugin is activated.**
+
+# Writing an Advanced Plugin
+
+An advanced plugin is one that is built as a self project and handles a version, tests, documentation, dev scripts and so on.
```
AnnaAleksandrowicz: @dan-ziv - what do you mean by "self project"?
@OrenMe - I know we want to keep the docs short, but in this case, I think we have to live with a longer document or split this into a series of pages, which may be a bit confusing. WDYT?

OrenMe: maybe `standalone github repo` is more suitable.
this is one of the more technical oriented docs so we can keep it as is

PR #85, opened 2018-01-31
fix(FEC-7815): Support playManifest redirects for external streams

odedhutzler:

Added jsonp configuration and a helper module to configure it in the player.
the configuration has 2 fields:
1. the jsonp handler function.
2. should we use the jsonp plugin.

### CheckLists

- [ ] Docs have been updated


src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+    parser.href = uri;
+    return {
+      'hostname': parser.hostname,
+      'uri': uri
+    }
+  };
+
+  const uriHost = getParsedUri(uri).hostname;
+  const flavorUriHost = getParsedUri(data.flavors[0].url).hostname
+  if (data.flavors.length === 1 && uriHost !== flavorUriHost) {
```
itaykinnrot: @odedhutzler please document this part inline - since this logic is not very clear at first

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+/**
+ * returns if we should use our jsonp http plugin
+ * @param {Object} config - configuration relevant to jsonp
+ * @returns {boolean} should or not use jsonp requests on manifests
+ */
+function shouldUseJsonp(config: Object): boolean{
+  if ((config && config.useJsonp) || Env.browser.name.includes("IE") || Env.browser.name.includes("Edge")){
```
itaykinnrot: @odedhutzler we already know that Panasonic browser need it too - please add

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * add jsonp configuration to the general config
+ * @param {Object} config - configuration relevant to jsonp
+ * @returns {void}
+ */
+function configureJsonp(config: Object): void{
+  config.playback = config.playback || {};
+  config.playback.options = config.playback.options || {};
+  config.playback.options.html5 = config.playback.options.html5 || {};
```
itaykinnrot: @odedhutzler we have a function for that please use it

src/common/utils/jsonp-helper.js
OrenMe: can we do better then browser sniffing?

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+  return false;
+}
+
+/**
+ * add jsonp configuration to the general config
+ * @param {Object} config - configuration relevant to jsonp
+ * @returns {void}
+ */
+function configureJsonp(config: Object): void{
+  config.playback = config.playback || {};
```
OrenMe: We should align this as general adapter config and not via direct and specific HLS or DASH settings.
try to generalise the setting.

src/kaltura-player.js
```diff
@@ -22,6 +23,7 @@ export default class KalturaPlayer {
   constructor(options: KalturaPlayerOptionsObject) {
     this._player = loadPlayer(options.player);
     this._logger = getLogger('KalturaPlayer' + Utils.Generator.uniqueId(5));
+    configureJsonp(this._player._config);
```
OrenMe: move to default options `getDefaultOptions?`

dan-ziv: 1. You are accessing a private member - you can't do it.
2. If you want to manipulate the configuration of the player you need to move the `configureJsonp` to the setup workflow, before the kaltura player has been created.

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+  return false;
+}
+
+/**
+ * add jsonp configuration to the general config
+ * @param {Object} config - configuration relevant to jsonp
+ * @returns {void}
+ */
+function configureJsonp(config: Object): void{
```
dan-ziv: I don't think `configureJsonp` is a good name, this is like to call a method `configureXHR`. JSONP is a generic method for sending JSON data. Try to think what is the real purpose here (redirect for external streams, for example).

dan-ziv: beautify

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+  config.playback = config.playback || {};
+  config.playback.options = config.playback.options || {};
+  config.playback.options.html5 = config.playback.options.html5 || {};
+  if (shouldUseJsonp(config.playback.options.html5.dash)){
+    config.playback.options.html5.dash = config.playback.options.html5.dash || {};
+    config.playback.options.html5.dash.useJsonp = true;
+    config.playback.options.html5.dash.callback = getDirectManfiestUri;
```
dan-ziv: Too generic names for configurations (see my suggestion bellow).

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {string} returns the direct uri
+ */
+function getDirectManfiestUri(data: Object, uri: string): string {
+  const getParsedUri = uri => {
+    const parser = document.createElement('a');
+    parser.href = uri;
+    return {
+      'hostname': parser.hostname,
+      'uri': uri
+    }
```
dan-ziv: missing semicolon

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * jsonp callback function, returns the direct manifest uri
+ * @param {Object} data - the json object that returns from the server
+ * @param {string} uri - original request uri
+ * @returns {string} returns the direct uri
+ */
+function getDirectManfiestUri(data: Object, uri: string): string {
+  const getParsedUri = uri => {
+    const parser = document.createElement('a');
```
dan-ziv: parser? looks like an element

src/common/utils/jsonp-helper.js
dan-ziv: Why to return an object if u use only the hostname field outside?

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+    parser.href = uri;
+    return {
+      'hostname': parser.hostname,
+      'uri': uri
+    }
+  };
+
+  const uriHost = getParsedUri(uri).hostname;
+  const flavorUriHost = getParsedUri(data.flavors[0].url).hostname
```
dan-ziv: unsafe access to object` data.flavors[0].url`. check for existence of fields.

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,62 @@
...
+function configureJsonp(config: Object): void{
+  config.playback = config.playback || {};
+  config.playback.options = config.playback.options || {};
+  config.playback.options.html5 = config.playback.options.html5 || {};
+  if (shouldUseJsonp(config.playback.options.html5.dash)){
+    config.playback.options.html5.dash = config.playback.options.html5.dash || {};
+    config.playback.options.html5.dash.useJsonp = true;
+    config.playback.options.html5.dash.callback = getDirectManfiestUri;
+  }
+  if (shouldUseJsonp(config.playback.options.html5.hls)){
```
dan-ziv: why separate config for hls and dash?
I don't even think its an engine dependent.
```js
{
playback: {
  options: {
    tryRedirectForExternalStreams: true,
    redirectForExternalStreamsHandler: () => {....}
    html5: {
    }
   }
 }
}
```  
Not even sure it should be under options, need to discuss the placement of this anyway.

src/common/utils/jsonp-helper.js
dan-ziv: beautify

src/common/utils/jsonp-helper.js
odedhutzler: it's purpose is to parse the url

src/common/utils/jsonp-helper.js
odedhutzler: k

src/common/utils/jsonp-helper.js
odedhutzler: k

src/common/utils/jsonp-helper.js
odedhutzler: i agree, though i must add the handler to the dash/hls config

dan-ziv: why?
in the `createAdapter` method of his adapter you get the whole player config.

src/common/utils/jsonp-helper.js
odedhutzler: @OrenMe I can change it to more configured... like
config.playback.options.adapters.forceRedriectBrowsers.includes(Env.browser.name) // "IE,Edge"

src/common/utils/jsonp-helper.js
odedhutzler: thought it gets the specific config, but you are right, i will change it.

src/common/utils/jsonp-helper.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * add jsonp configuration to the general config
+ * @param {Object} config - configuration relevant to jsonp
+ * @returns {void}
+ */
+function configureExternalStreamRedirect(config: Object): void {
+  config.playback.options = config.playback.options || {};
+  let adapters = config.playback.options.adapters = config.playback.options.adapters || {};
+  if (typeof adapters.forceRedirectExternalStreams !== "boolean") {
```
dan-ziv: @OrenMe 
We should discuss those configuration names and placement.
I think it should be forceRedirect**To**ExternalStreams and under sources config (it related directly to sources)

PR #126, opened 2018-05-23
fix(FEC-8240): support legacy config on configure API

OrenMe:

The legacy config transformation is only applied on the initial setup call, so need to apply it also for the configure API.

### CheckLists

- [ ] Docs have been updated


yairans:
@OrenMe LGTM but please add a test (just copy the Dan's test for change media)

OrenMe:
@yairans totally agree, unfortunately we do not have any tests for the kaltura-player exposed API yet(configure, loadMedia and setMedia)

PR #129, opened 2018-06-11
fix(FEC-8298): SOURCE_SELECTED event triggered before the UI ready when the source provided manually

yairans:

postpone the `sources` config after the ui initialization  

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/kaltura-player.js
+++ b/src/kaltura-player.js
@@ -40,13 +40,16 @@ class KalturaPlayer extends FakeEventTarget {
   constructor(options: KPOptionsObject) {
     super();
     this._eventManager = new EventManager();
-    this._localPlayer = loadPlayer(options);
+    const {sources} = options;
+    const noSourcesOptions = Utils.Object.mergeDeep({}, options, {sources: null});
+    this._localPlayer = loadPlayer(noSourcesOptions);
     this._logger = getLogger('KalturaPlayer' + Utils.Generator.uniqueId(5));
     this._uiWrapper = new UIWrapper(this, options);
     this._provider = new Provider(options.provider, __VERSION__);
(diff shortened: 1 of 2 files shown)


src/kaltura-player.js
```diff
@@ -40,13 +40,16 @@ class KalturaPlayer extends FakeEventTarget {
   constructor(options: KPOptionsObject) {
     super();
     this._eventManager = new EventManager();
-    this._localPlayer = loadPlayer(options);
+    const {sources} = options;
```
OrenMe: what happens if `sources` doesn't exist? what is the value?

src/kaltura-player.js
```diff
@@ -40,13 +40,16 @@ class KalturaPlayer extends FakeEventTarget {
...
+    const noSourcesOptions = Utils.Object.mergeDeep({}, options, {sources: null});
+    this._localPlayer = loadPlayer(noSourcesOptions);
     this._logger = getLogger('KalturaPlayer' + Utils.Generator.uniqueId(5));
     this._uiWrapper = new UIWrapper(this, options);
     this._provider = new Provider(options.provider, __VERSION__);
     this._playlistManager = new PlaylistManager(this, options);
     this._playlistManager.configure(options.playlist);
     Object.values(CoreEventType).forEach(coreEvent => this._eventManager.listen(this._localPlayer, coreEvent, e => this.dispatchEvent(e)));
+    this._localPlayer.configure({sources});
```
OrenMe: check if sources exist, don't call if it's null/undefined

src/kaltura-player.js
yairans: we have default sources

PR #130, opened 2018-06-14
feat(FEC-8046): 360 support

yairans:

* force inBrowserFullscreenForIOS when 'vr' plugin is set
* set `vrStereo` ui component when 'vr' plugin is set
* pass the target dom element to 'vr' plugin

### CheckLists

- [x] Docs have been updated


src/common/ui-wrapper.js
```diff
@@ -38,6 +38,14 @@ class UIWrapper {
     this.setConfig(Utils.Object.mergeDeep({}, previewThumbnailConfig, seekbarConfig), 'seekbar');
   }
 
+  setFullscreenConfig(config: ProviderMediaConfigObject): void {
+    if (this._disabled) return;
+    if (Utils.Object.getPropertyPath(config, 'plugins.vr')) {
```
yairans: maybe better to check the plugin does exists via `getPlugin('vr')`

src/kaltura-player.js
```diff
@@ -77,4 +79,16 @@ export default class KalturaPlayer {
       set: undefined
     };
   }
+
+  _addBindings(): void {
```
dan-ziv: This is massive and too specific referring between kaltura player and a plugin
If the problem is that the plugin cannot listen to the UI events the solution is to move the plugin manager to the kaltura player level and not bind the kaltura player to a specific plugin

dan-ziv: one more thing is - if we will run kaltura player in chromeless mode (`{ui: {disable: true}}`) the vr navigation won't work.

src/common/plugins/plugins-config.js
```diff
@@ -12,7 +12,8 @@ function evaluatePluginsConfig(options: KalturaPlayerOptionsObject): void {
   if (options.plugins) {
     const dataModel: Object = {
       pVersion: __VERSION__,
-      pName: __NAME__
+      pName: __NAME__,
+      target: options.targetId
```
OrenMe: target is too generic term, suggest to change to domRootElementId

src/common/ui-wrapper.js
```diff
@@ -38,6 +38,25 @@ class UIWrapper {
...
+    }
+  }
+
+  _setFullscreenConfig(): void {
+    const fullscreenConfig = Utils.Object.getPropertyPath(this._uiManager, 'config.components.fullscreen');
+    this.setConfig(Utils.Object.mergeDeep({}, {inBrowserFullscreenForIOS: true}, fullscreenConfig), 'fullscreen');
+  }
+
+  _setStereoConfig(config: KalturaPlayerOptionsObject): void {
+    if (config.plugins.vr.toggleStereo || (Env.device.type && config.plugins.vr.toggleStereo !== false)) {
```
OrenMe: not sure I understand this condition

src/kaltura-player.js
```diff
@@ -23,12 +23,14 @@ export default class KalturaPlayer {
     this._logger = getLogger('KalturaPlayer' + Utils.Generator.uniqueId(5));
     this._uiWrapper = new UIWrapper(this._player, options.ui);
     this._provider = new Provider(options.provider, __VERSION__);
+    this._uiWrapper.handleVr(options);
```
OrenMe: why not handle this in the uiWrapper?

src/kaltura-player.js
OrenMe: @dan-ziv, @yairans  we will move this in to the player and then the UI will call `player.toggleStereoMode()` directly

src/common/ui-wrapper.js
yairans: @OrenMe  `toggleStereo` is `false` by default for desktop and `true` for mobile. is it clear?

src/common/plugins/plugins-config.js
yairans: done

src/kaltura-player.js
yairans: done

src/kaltura-player.js
yairans: done

PR #139, opened 2018-07-01
feat(FEC-8364): support adding evaluated expressions also on player setup and configure API

OrenMe:

When calling player.setup or player.configure we need to:
1. Evaluate new expressions that are passed with tokens
2. Add new evaluation tokens to the list of evaluated tokens so they can be re-evaluated on each media change.

This feature will create a plugin token config store.
On each call to evaluate the list of new expressions will be analyzed, and if it contains evaluation tokens these tokens will be analyzed and also be saved in the config store so in next call to configure, loadMedia or setMedia the player will run the evaluation process on them.

**[Still needs test and docs]**

### CheckLists

- [ ] Docs have been updated


src/common/plugins/plugins-config.js
```diff
@@ -1,43 +1,95 @@
...
 import {Utils} from 'playkit-js'
 
+/**
+ * returns weather value is evaluated
+ * @param {*} value - the value to be checked
+ * @returns {boolean} - value is evaluated
+ */
+const valueIsEvaluated= (value: any): boolean =>
+  (typeof value === "number" || typeof value === "string" || typeof value === "boolean") &&
+  !templateRegex.test(value.toString());
```
odedhutzler: maybe isValueEvaluated

src/common/plugins/plugins-config.js
```diff
@@ -1,43 +1,95 @@
...
+        return product;
+      },
+      {}
+    );
+
+/**
+ * returns the data model for evaluating evaluation tokens
+ * @param {KalturaPlayerOptionsObject} options - the kaltura player options object
+ * @returns {Object} - data model
+ */
```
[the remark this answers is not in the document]
OrenMe: It could be nice but as this is an open object from the beginning it is basically `{[name: string]: Object}` so no real gain here TBH

src/common/plugins/plugins-config.js
OrenMe: done

PR #144, opened 2018-07-05
fix(FEC-8382): stereo icon displayed on iPhone, even if VR plugin disabled

yairans:

check vr plugin exists and not disable before handling it
 
### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/ui-wrapper.js
+++ b/src/common/ui-wrapper.js
@@ -29,17 +29,16 @@ class UIWrapper {
   }
 
   resetErrorConfig(mediaInfo: ProviderMediaInfoObject): void {
+    if (this._disabled) return;
     this._setErrorPresetConfig(mediaInfo);
     this._resetErrorState();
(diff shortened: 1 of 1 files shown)


src/common/ui-wrapper.js
```diff
@@ -50,8 +49,8 @@ class UIWrapper {
     this.setConfig(Utils.Object.mergeDeep({}, previewThumbnailConfig, seekbarConfig), 'seekbar');
   }
 
-  _handleVr(config: ?PKPluginsConfigObject): void {
-    if (config) {
+  _handleVr(config: PKPluginsConfigObject = {}): void {
+    if (config.vr && !config.vr.disable) {
```
yairans: Actually, I would like to be sure the plugin is active (maybe its bundle non exist at all?) 
But this info is unavailable in kalturaPlayer yet.

PR #145, opened 2018-07-08
fix(FEC-8386): Kava analytics retrieved by default on OTT

yairans:

Please add a detailed description of the change, weather it's an enhancement or a bugfix.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/kaltura-player.js
+++ b/src/kaltura-player.js
@@ -59,7 +59,7 @@ export default class KalturaPlayer {
     const playerConfig = Utils.Object.copyDeep(mediaConfig);
     Utils.Object.mergeDeep(playerConfig.sources, this._player.config.sources);
     Utils.Object.mergeDeep(playerConfig.session, this._player.config.session);
-    Object.keys(this._player.config.plugins).forEach(name => {playerConfig.plugins[name] = {}});
+    Object.keys(this._player.config.plugins).forEach(name => {playerConfig.plugins[name] = {disable: this._player.config.plugins[name].disable}});
     addKalturaPoster(playerConfig.sources, mediaConfig.sources, this._player.dimensions);
     addKalturaParams(this._player, playerConfig);
     this._uiWrapper.setSeekbarConfig(mediaConfig, this._player.config.ui);


src/kaltura-player.js
```diff
@@ -59,7 +59,7 @@ export default class KalturaPlayer {
     const playerConfig = Utils.Object.copyDeep(mediaConfig);
     Utils.Object.mergeDeep(playerConfig.sources, this._player.config.sources);
     Utils.Object.mergeDeep(playerConfig.session, this._player.config.session);
-    Object.keys(this._player.config.plugins).forEach(name => {playerConfig.plugins[name] = {}});
+    Object.keys(this._player.config.plugins).forEach(name => {playerConfig.plugins[name] = {disable: this._player.config.plugins[name].disable}});
```
OrenMe: This seems like a weakness in the way we set disable.
The disable should be set on the initial call to `setup`

yairans: @OrenMe we did it on the initial call to `setup` but override it on `setMedia`

OrenMe: So instead of passing it again just filter the disabled plugins out before passing them to be evaluated in the setMedia.
In any case this is a bug in our pluginManager cause if a plugin was set with disable true and then configure was called again with that plugin name without setting disable false explicitly it will be activated.
I think we should add some sort of API for plugins, plugins namespace with getActive, getRegistered, enable, disable

yairans: @OrenMe the plugin disabling is handled in the plugin manager. why to handle it also here?

OrenMe: @yairans that’s exactly what I’m saying,but  bug is in Playkit plugin manager.
If plugin is set as disable then setting any config on it in Playkit shouldn’t result in uniting it.
The problem is we don’t protect for this.
Reading the playkit  config here before passing config back to player is what we had in the beginning and it led to trouble as there’s a hard time understanding from where config came from.
The issue now is that we pass disable: true in setup call and then we pass pluginConfig again on setMedia. This new config doesn’t contain the disable: false in it but Playkit, that gets this config, doesn’t care that plugin is disabled, because we don’t save this anywhere, and goes to initOrConfigure method and sees that plugin is not init(because we had disable: true before) and now it init it, although it shouldn’t cause disable: false was NEVER set after disable: true was.

PR #151, opened 2018-08-14
fix(FEC-8474): Playback starts muted on change media - Ios

yairans:

Please add a detailed description of the change, weather it's an enhancement or a bugfix.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/utils/setup-helpers.js
+++ b/src/common/utils/setup-helpers.js
@@ -1,6 +1,6 @@
 // @flow
 import {setDefaultAnalyticsPlugin} from 'player-defaults';
-import {Env, TextStyle, Utils} from 'playkit-js';
+import {Env, TextStyle, Utils, getCapabilities} from 'playkit-js';
 import {ValidationErrorType} from './validation-error';
 import StorageManager from '../storage/storage-manager';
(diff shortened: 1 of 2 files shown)


src/common/utils/setup-helpers.js
```diff
@@ -104,6 +104,22 @@ function setStorageTextStyle(player: Player): void {
...
+ * @returns {void}
+ */
+function attachToFirstClick(player: Player): void {
+  if (isIos()) {
+    const onUIClicked = () => {
+      player.removeEventListener(player.Event.UI.UI_CLICKED, onUIClicked);
+      // Trigger to load the auto play testing video tag
+      getCapabilities();
+    };
+    player.addEventListener(player.Event.UI.UI_CLICKED, onUIClicked);
```
yairans: This change assumes that by the first `UI_CLICKED` the main video tag is always can auto play.
But this assumption is **wrong!!!**. 
Consider a case that the first entry request is failed, and the player is in an error state.
By click on the error overlay (for retry for example) we get `UI_CLICKED` but the main video tag is still **cannot** auto play, as it has never played.

PR #168, opened 2018-11-08
fix(FEC-8588): large play button displayed for a sec before the next item is playing

yairans:

use `autoplay` config for the second entry and onwards 

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/playlist/playlist-manager.js
+++ b/src/common/playlist/playlist-manager.js
@@ -1,6 +1,6 @@
 // @flow
 import {KalturaPlayer} from '../../kaltura-player';
-import {FakeEvent, Utils} from '@playkit-js/playkit-js';
+import {FakeEvent, Utils, EventManager} from '@playkit-js/playkit-js';
 import {PlaylistEventType} from './playlist-event-type';
 import getLogger from '../utils/logger';
 import {Playlist} from './playlist';
@@ -17,6 +17,7 @@ class PlaylistManager {
 
   constructor(player: KalturaPlayer, options: KPOptionsObject) {
     this._player = player;
+    this._eventManager = new EventManager();
     this._playlist = new Playlist();
     this._options = {autoContinue: true};
     this._countdown = {duration: 10, showing: true};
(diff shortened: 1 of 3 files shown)


yairans:
@dan-ziv the large play button just raised the issue, but the solution is correct any way. play next means autoplay true.

dan-ziv:
@yairans No, play next means **autoContinue** true. The autoplay is separate configuration which is not related to the playlist (In playlist context it effected only the first entry). There's no issue here from the player side.

yairans:
@dan-ziv the core player doesn't know about autoContinue config and playlist flow at all.
from its side play next means load an entry and play it automatically

dan-ziv:
The core player isn't related to this. The kaltura player manage the playlist logic and in the same way that the core play manage autoplay (by internally calling play()) the kaltura player should manage autoContinue. The above issue is a pure UI issue.

src/common/playlist/playlist-manager.js
```diff
@@ -34,14 +36,18 @@ class PlaylistManager {
         this._player.dispatchEvent(new FakeEvent(PlaylistEventType.PLAYLIST_LOADED, {playlist: this._playlist}));
         const next = this._playlist.next;
         if (next.item) {
-          this._setItem(next.item, next.index);
+          this._setItem(next.item, next.index).then(() => {
+            this._eventManager.listenOnce(this._player, PlaylistEventType.PLAYLIST_ITEM_CHANGED, () => {
+              this._player.configure({playback: {autoplay: this._options.autoContinue}});
```
OrenMe: wouldn't that cause each call to configure to do this? so calling configure multiple times will cause this config to fire.

yairans: fixed. 10x.
new it's only when new sources configured

yairans:
@dan-ziv as approved by @itaykinnrot and @OrenMe any play next flow (even by user gesture) consider as auto play. So this fix is correct also without the UI issue.

PR #194, opened 2018-12-26
fix(FEC-8791): When changing volume in local mode it's not showing the correct rate/state when returning to local mode after casting

dan-ziv:

Set the storage config values when disconnecting from the remote player.

### CheckLists

- [ ] Docs have been updated


src/common/cast/remote-control.js
```diff
@@ -216,10 +217,14 @@ function reconstructPlayerComponents(snapshot: PlayerSnapshot): void {
 
 function configurePlayback(playbackConfig: Object): void {
   const {autoplay, startTime} = playbackConfig;
+  const storageConfig = StorageManager.getStorageConfig();
+  const {muted, volume} = storageConfig;
```
OrenMe: what happens if muted and/or volume are undefined from storage manger?

dan-ziv: nothing will happens

PR #207, opened 2019-01-15
chore: fixed playkit-js-providers tag name

rkreich:
When including `kaltura-player-js` in 3rd party projects that use NPM, dependencies installation will fail due to missing tag. `2.14.0` instead of `v2.14.0`.


OrenMe:
fixed in another commit, tnx!

PR #221, opened 2019-02-19
fix(FEC-8901): duplication of ClosedCaptions on iOS on none Safari browser

shalommeoded:

Changed the `if` condition of `checkNativeTextTracksSupport` function, to allow other browsers in iOS to set it to `true` by default.

From `isSafari()` -> to `isSafari() || isIos()`

- Tested the matter with iPad in Chrome, Safari, and FireFox.

### CheckLists

- [ ] Docs have been updated


src/common/utils/setup-helpers.js
```diff
@@ -306,7 +306,7 @@ function configureDelayAdsInitialization(options: KPOptionsObject): void {
  * @returns {void}
  */
 function checkNativeTextTracksSupport(options: KPOptionsObject): void {
-  if (isSafari()) {
+  if (isIos()) {
```
yairans:
```suggestion
  if (isSafari() || isIos()) {
```
do include safari desktop as well

PR #247, opened 2019-06-18
feat(FEC-9023): more logger options

Yuvalke:

add configuration logger options.
handler to output logs with customize function.
option to remove playerVersion from log.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/utils/setup-helpers.js
+++ b/src/common/utils/setup-helpers.js
@@ -196,7 +196,6 @@ function setLogOptions(options: KPOptionsObject): void {
 
   // $FlowFixMe
   options.ui.log.level = options.provider.log.level = logLevelObj.name;
-
   _setLogLevel(logLevelObj);
 }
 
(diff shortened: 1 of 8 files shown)


src/common/utils/setup-helpers.js
```diff
@@ -175,15 +175,27 @@ function isDebugMode(): boolean {
...
+
+  if (options.log && typeof options.log.handler === 'function') {
+    setLogHandler(options.log.handler);
+    // $FlowFixMe
+    options.ui.log.handler = options.provider.log.handler = options.log.handler;
+  }
+
   let logLevelObj: LogLevelObject = LogLevel.ERROR;
-  if (isDebugMode()) {
+  if (options.log && isDebugMode()) {
```
yairans: 1. `options.log` must be. you just create it a few lines above 
2. `isDebugMode` got from the url so it must to be applied. with no `options.log` condition

yairans: does flow require it?

Yuvalke: Flow require

PR #253, opened 2019-07-08
Fec 8696 contrib presets

esakal:


src/common/ui-wrapper.js
```diff
@@ -10,6 +10,7 @@ class UIWrapper {
 
   constructor(player: KalturaPlayer, options: KPOptionsObject) {
     const config: UIOptionsObject = options.ui;
+    (config.presetComponents || (config.presetComponents = [])).push(...player._localPlayer.presetComponents);
```
esakal: @OrenMe the code above assume that `presetComponents` can be provided in two ways:
1. from each plugin using the new interface
2. statically from the page that configured the ui

do you want to keep both or only the first one?

yairans: And if we want to keep both, what's about merging them?

yairans:
```suggestion
    (config.presetComponents || (config.presetComponents = [])).concat(player._localPlayer.presetComponents);
```

OrenMe: Keep both, what happens if app provides same component as plugin, let's assume different use cases, like app wanting to override default plugin component.
A rule of thumb in player, app(or user) always gets precedence over default config.

esakal: done

esakal:
not relevant anymore

PR #254, opened 2019-07-10
feat(FEC-9227): support reInit of MSE

Yuvalke:

Add configuration to handle ads with MSE

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/utils/setup-helpers.js
+++ b/src/common/utils/setup-helpers.js
@@ -549,7 +549,7 @@ function isIos(): boolean {
  * @returns {boolean} - if browser is in LG TV
  */
 function isSmartTv(): boolean {
-  return Env.isSmartTV;
+  return Env.os.name.toLowerCase() === 'web0s' || Env.os.name.toLowerCase() === 'tizen';
 }
 
 /**
--- a/src/common/utils/setup-helpers.js
+++ b/src/common/utils/setup-helpers.js
@@ -339,14 +339,10 @@ function _configureAdsWithMSE(options: KPOptionsObject): void {
   const preferNativeDash = !!Utils.Object.getPropertyPath(options, 'playback.preferNative.dash');
   if ((!preferNativeHls || !preferNativeDash) && options.plugins && options.plugins.ima) {
     const playAdsWithMSE = Utils.Object.getPropertyPath(options, 'playback.playAdsWithMSE');
-    const reloadSourceOnComplete = Utils.Object.getPropertyPath(options, 'plugins.ima.reloadSourceOnComplete');
     const disableMediaPreload = Utils.Object.getPropertyPath(options, 'plugins.ima.disableMediaPreload');
 
     if (typeof playAdsWithMSE !== 'boolean') {
       options = Utils.Object.createPropertyPath(options, 'playback.playAdsWithMSE', true);
-    }
-    if (typeof reloadSourceOnComplete !== 'boolean') {
-      options = Utils.Object.createPropertyPath(options, 'plugins.ima.reloadSourceOnComplete', false);
     }
     if (typeof disableMediaPreload !== 'boolean') {
       options = Utils.Object.createPropertyPath(options, 'plugins.ima.disableMediaPreload', true);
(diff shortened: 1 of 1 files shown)


src/common/utils/setup-helpers.js
```diff
@@ -328,18 +328,48 @@ function checkNativeTextTracksSupport(options: KPOptionsObject): void {
...
  * @returns {void}
  */
 function configureLGTVDefaultOptions(options: KPOptionsObject): void {
-  if (isLGTV()) {
+  if (isSmartTv()) {
     const preferNativeHls = Utils.Object.getPropertyPath(options, 'playback.preferNative.hls');
+    const preferNativeDash = Utils.Object.getPropertyPath(options, 'playback.preferNative.dash');
     if (typeof preferNativeHls !== 'boolean') {
-      options = Utils.Object.createPropertyPath(options, 'playback.preferNative.hls', true);
+      options = Utils.Object.createPropertyPath(options, 'playback.preferNative.hls', false);
```
yairans: this is the default config. you don't need config it explicitly

src/common/utils/setup-helpers.js
```diff
@@ -328,18 +328,48 @@ function checkNativeTextTracksSupport(options: KPOptionsObject): void {
...
+    }
+    if (typeof preferNativeDash !== 'boolean') {
+      options = Utils.Object.createPropertyPath(options, 'playback.preferNative.dash', false);
```
yairans: ditto

Yuvalke: Right, but If someone will handle the opposite after that it'll not work..
What do you think?

yairans: if someone will change the player default he will break the player, we have a lot of assumptions on the player defaults

src/common/utils/setup-helpers.js
```diff
@@ -328,18 +328,36 @@ function checkNativeTextTracksSupport(options: KPOptionsObject): void {
...
+/**
+ * Sets config option for Ads with MSE
+ * @private
+ * @param {KPOptionsObject} options - kaltura player options
+ * @returns {void}
+ */
+function _configureAdsWithMSE(options: KPOptionsObject): void {
+  const preferNativeHls = !!Utils.Object.getPropertyPath(options, 'playback.preferNative.hls');
+  const preferNativeDash = !!Utils.Object.getPropertyPath(options, 'playback.preferNative.dash');
+  if ((!preferNativeHls || !preferNativeDash) && options.plugins && options.plugins.ima) {
```
yairans: 1. what's about `preferNativeHls: true` but `preferNativeDash:false` (very common)
2. you can remove the native condition as you moved the attach/detach to the adapters
3. if you verify ima you have to add 1. `ima.disable:false` 2.bumper 
4. whats about dai, you can't detach/attach because the ad is part of the stream

src/common/utils/setup-helpers.js
```diff
@@ -328,18 +328,41 @@ function checkNativeTextTracksSupport(options: KPOptionsObject): void {
...
+    if (options.plugins && options.plugins.imadai && !options.plugins.imadai.disable) {
+      options = Utils.Object.createPropertyPath(options, 'playback.playAdsWithMSE', false);
+    } else {
+      options = Utils.Object.createPropertyPath(options, 'playback.playAdsWithMSE', true);
+    }
+  }
+  const disableMediaPreloadIma = Utils.Object.getPropertyPath(options, 'plugins.ima.disableMediaPreload');
+  const disableMediaPreloadBumper = Utils.Object.getPropertyPath(options, 'plugins.bumper.disableMediaPreload');
+
+  if (options.plugins && options.plugins.ima && typeof disableMediaPreloadIma !== 'boolean') {
```
yairans:
```suggestion
  if (typeof disableMediaPreloadIma !== 'boolean') {
```
`createPropertyPath` already validates the path , same for bumper.

Yuvalke: I don't want to validate it...
If IMA doesn't exist it'll create the plugin config with disablemediapreload only, and it's not right behaviour

PR #299, opened 2020-01-23
feat(FEC-8998): add url encoded for referrer token

RoyBregman:

add new property "encodedReferrer" to plugins config model which uses a new method getEncodedReferrer in kaltura-params which encodes the getReferrer

solves FEC-8998

### CheckLists

- [X] new unit / functional tests have been added (whenever applicable)
- [ ] Docs have been updated

### DIFF
--- a/src/common/utils/kaltura-params.js
+++ b/src/common/utils/kaltura-params.js
@@ -94,15 +94,6 @@ function getReferrer(): string {
 }
 
 /**
- * @return {string} - The referrer after URIComponent encoded
- * @private
- */
-function getEncodedReferrer(): string {
-  const referrer = getReferrer();
-  return typeof referrer === 'string' ? encodeURIComponent(referrer) : referrer;
-}
-
-/**
  * @param {PKMediaSourceObject} source - source
  * @return {void}
  * @private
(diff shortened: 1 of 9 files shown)


src/common/utils/kaltura-params.js
```diff
@@ -93,6 +93,15 @@ function getReferrer(): string {
   return referrer;
 }
 
+/**
+ * @return {string} - The referrer after URIComponent encoded
+ * @private
+ */
+function getEncodedReferrer(): string {
```
yairans: if it's not relevant to `kaltura-params.js` move it to `plugins-config.js`.

RoyBregman: fixed

src/common/utils/kaltura-params.js
```diff
@@ -93,6 +93,15 @@ function getReferrer(): string {
...
+  const referrer = getReferrer();
+  return typeof referrer === 'string' ? encodeURIComponent(referrer) : referrer;
```
OrenMe: do we `typeof` this because of flow?

OrenMe: when can it be undefined or not a string?

RoyBregman: in reality according to the current implementation of getReferrer it doesn't seem possible. Still i think it is good to be sure and wrap it in case getReferrer returns undefined in some weird scenario (for example an old browser or something)

OrenMe: not sure I follow - do you know old browsers return undefined? according to MDN it is always a string: https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

RoyBregman: removed check

src/common/utils/kaltura-params.js
OrenMe: @yairans please check approve

PR #326, opened 2020-06-15
fix(FEC-10122): destroy cleanup process

Yuvalke:

static class added listeners to the player instance which weren't cleared on destroy.
Add destroy to playlist manager.

### CheckLists

- [ ] Docs have been updated


src/common/playlist/playlist-manager.js
```diff
@@ -282,6 +282,12 @@ class PlaylistManager {
     }
     return Promise.reject();
   }
+
+  destroy(): void {
+    delete this._player;
+    delete this._playlist;
```
yairans, line 288: not need to delete them. we don't do like this in any file

Yuvalke: Why? set to null is not acceptable by flow, I think the correct way is to delete the ref prop from class on destroy

OrenMe: @Yuvalke are you profiling and seeing that on destroy the player keeps reference?
The correct way is to set null from the app side, e.g. 
```javascript
const player = KalturaPlayer.setup(); 
//do stuff
player.destroy();
player = null;
```

PR #329, opened 2020-06-23
feat(FEC-9649): loading external CSS

RoyBregman:

In UIWrapper constructor we check if the ui config has a css url provided.
If so we load it and append it to the doc header

solves FEC-9649

### CheckLists

- [ ] Docs have been updated


src/common/ui-wrapper.js
```diff
@@ -58,6 +66,19 @@ class UIWrapper {
...
   }
 
+  _handleExternalCSS(config: UIOptionsObject = {}): void {
+    if (config.css) {
+      Utils.Dom.loadStyleSheetAsync(config.css).then(
+        () => {
+          logger.debug(`external css was loaded successfully`);
+        },
+        () => {
+          logger.error(`external css failed to load`);
```
dan-ziv, line 76: Is it needed to maybe add some small retry logic if the loading has been failed?

RoyBregman: As we discussed this morning this is just a link with a source. If want to add retry mechanism we need to make something generic for all loadings

PR #330, opened 2020-06-28
fix: remove listeners which added by static class on destroy

Yuvalke:

static class added listeners which weren't cleared on destroy.

### CheckLists

- [ ] Docs have been updated


Yuvalke:
> no need to manage with event manager here as the player is managing it as well and we don't keep any reference here.

It's not true, cause we have a new registration to the player instance and it'll never be cleared. I keep the instance in of vent manager to destroy the listeners in destroy.

PR #349, opened 2020-08-13
fix(FEC-10417): playlist by sources stuck after press Play button when set IMA or bumper plugins

dan-ziv:

When playlist configured with sources the play flow start from the `configure` of the playlist manager.
In that point, no listener and handlers attached on the player, so nothing happens when playlist item is set.
The solution is the bind all the listeners first, and then configure the playlist manager.

Solves FEC-10417

### CheckLists

- [ ] Docs have been updated


OrenMe:
@yairans please check why we didn't see this in tests(do we have for this kind of thing?)

yairans:
@dan-ziv please add a unit test as discussed

yairans (approved):

PR #353, opened 2020-09-01
fix(FEC-10455): incorrect order in reset and destroy process

Yuvalke:

Issue: player.destroy() sending invalid 'stop' bookmark with position val 0
Solution: reorder the components destroy/reset since some of them depend on playkit.

Solve: FEC-10455.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/kaltura-player.js
+++ b/src/kaltura-player.js
@@ -238,9 +238,9 @@ class KalturaPlayer extends FakeEventTarget {
       this._reset = true;
       this._firstPlay = true;
       this._playbackStart = false;
-      this._localPlayer.reset();
       this._uiWrapper.reset();
       this._pluginManager.reset();
+      this._localPlayer.reset();
     }
   }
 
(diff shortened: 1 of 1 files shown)


OrenMe (changes requested):
changing the order will most likely have implications.
I think moving the plugin manager to be before the local player is ok cause before the plugin manager changes it was almost first in playkit-js, but ui wrapper in reset I don't think so and for the destroy it might also be a drastic change.

In general I agree that all managers beside event need to be destroyed before player - but please check the implications so we don't have dependencies between components in destruction order.
@yairans please also review as you were the last one to touch this AFAICR

Yuvalke:
> changing the order will most likely have implications.
> I think moving the plugin manager to be before the local player is ok cause before the plugin manager changes it was almost first in playkit-js, but ui wrapper in reset I don't think so and for the destroy it might also be a drastic change.
> 
> In general I agree that all managers beside event need to be destroyed before player - but please check the implications so we don't have dependencies between components in destruction order.
> @yairans please also review as you were the last one to touch this AFAICR

@OrenMe I did check for Playlist Pugins and UI, all of them depend on playkit and UI and plugins used playkit on destroy/reset on the other hand playkit doesn't depend on none of them, same as each one in another so I think it's safe.

yairans (approved):

OrenMe (approved):
Ok, LGTM.

PR #356, opened 2020-09-03
fix(FEC-10468): PLAYBACK_START not fired on autoplay

yairans:

Move `PLAYBACK_START` back to playkit for autoplay flow
(The only reason we moved it to kaltura player is to be aligned with PLAYBACK_ENDED)

Revert part of commit 6217699

Solves FEC-10468

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
-    "@playkit-js/playkit-js": "0.63.0",
+    "@playkit-js/playkit-js": "0.64.0-canary.950da7a",
     "@playkit-js/playkit-js-dash": "1.19.1",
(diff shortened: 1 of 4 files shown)


Yuvalke (approved):

Yuvalke (changes requested):
could you move the tests to playkit?

yairans:
why move them? I want to be sure it dispatched by kaltura player

Yuvalke:
> why move them? I want to be sure it dispatched by kaltura player

NP so add tests for playkit cause if someone will change the logic in playkit, we should see the problem in playkit.

Yuvalke (dismissed):
I think the autoplay test should be added on playkit as well.

PR #358, opened 2020-09-13
feat(FEC-10405): Casting with Airplay

dan-ziv:

Configure airplay plugin when running on Safari.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/utils/setup-helpers.js
+++ b/src/common/utils/setup-helpers.js
@@ -292,6 +292,7 @@ function getDefaultOptions(options: PartialKPOptionsObject): KPOptionsObject {
   configureExternalStreamRedirect(defaultOptions);
   maybeSetFullScreenConfig(defaultOptions);
   maybeSetCapabilitiesForIos(defaultOptions);
+  maybeConfigureAirplay(defaultOptions);
   return defaultOptions;
 }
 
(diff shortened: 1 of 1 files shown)


src/common/utils/setup-helpers.js
```diff
@@ -638,6 +639,22 @@ function maybeSetCapabilitiesForIos(options: KPOptionsObject): void {
...
 }
 
+/**
+ * Configures airplay plugin in case we're running in Safari browser.
+ * @private
+ * @param {KPOptionsObject} options - kaltura player options
+ * @returns {void}
+ */
+function maybeConfigureAirplay(options: KPOptionsObject): void {
+  if (Env.isSafari) {
```
Yuvalke, line 649: Please add !Env.isSmartTV since LG2 recognize and safari browser

Yuvalke (changes requested):

src/common/utils/setup-helpers.js
OrenMe: so it makes more sense to check if ios or macOS then to check !smartTV in this context

OrenMe (commented):
@dan-ziv I think we don't need this if we are going with explicit plugin addition via studio, right?

src/common/utils/setup-helpers.js
dan-ziv: This PR is irrelevant since we will configure it via Studio

PR #369, opened 2020-11-03
docs(FEC-10387): add support for referenceId filter in provider

RoyBregman:

Updating docs to support media loading with referenceId instead of entryId

solves FEC-10387

### CheckLists

- [X] Docs have been updated


yairans:
@RoyBregman What's about analytics?
This field I believe should be passed on the plugin config using the evaluation mechanism.

RoyBregman:
> @RoyBregman What's about analytics?
> This field I believe should be passed on the plugin config using the evaluation mechanism.
> Then need to consider it in the analytics plugins

Awesome question, checked it now and it is okay. Uses the entryId from the response

yairans (dismissed):

OrenMe (changes requested):
change title to `docs` if this is just docs.
Need to explain order of importance between entry and reference id.

OrenMe (dismissed):

OrenMe (approved):

PR #372, opened 2020-11-12
feat(FEC-10233): plugin async loading support

RoyBregman:

Added a new getter "Ready" in BasePlugin which is resolved by default.
If a plugin wants to make sure load and play don't occur before some async event in the plugin he can override the getter and return a new promise which will be resolved accordingly.

A new Middleware "PluginReadinessMiddleware" was created and added to the beginning of the middlewares chain reaction to make sure all plugin promises are resolved before starting the load / play middleware chain reaction

resolves FEC-10233

### CheckLists

- [X] Docs have been updated


dan-ziv:
@RoyBregman  please run `prettier:fix`

src/common/plugins/plugin-readiness-middleware.js
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Load middleware handler.
+   * @param {Function} next - The load handler in the middleware chain.
+   * @returns {void}
+   * @memberof PluginReadinessMiddleware
+   */
+  load(next: Function): void {
+    this.checkNextSettle(0, next);
+  }
+  checkNextSettle(index, next: Function) {
```
OrenMe, line 25: is this an API? if not add _ to signal as private.

src/common/plugins/plugin-readiness-middleware.js
```diff
@@ -0,0 +1,54 @@
...
+    if (index < this._plugins.length) {
+      this.checkSettle(index, next);
+    } else {
+      this.callNext(next);
+    }
+  }
+  checkSettle(index: number, next: Function) {
```
OrenMe, line 32: is this an API? if not add _ to signal as private.

src/common/plugins/plugin-readiness-middleware.js
```diff
@@ -0,0 +1,54 @@
...
+    const readyPromise = this._plugins[index].ready ? this._plugins[index].ready : Promise.resolve();
```
OrenMe, line 33: As we discussed, I thought this is a good protection but in general any plugin using the Kaltura Player that implements this will have this base plugin so there shouldn't be any issue with the base plugin always returning `Promise.resolve()`

OrenMe (changes requested):

src/common/plugins/plugin-readiness-middleware.js
RoyBregman: This is for external plugins

src/common/plugins/plugin-readiness-middleware.js
RoyBregman: done

src/common/plugins/plugin-readiness-middleware.js
RoyBregman: done

docs/how-to-build-a-plugin.md
```diff
@@ -175,6 +175,14 @@ The player will call this method before destroying itself.
...
 The player will call this method before changing media.
 
+> #### `get ready()`
+>
+> Returns: `Promise<*>` - a Promise which is resolved when plugin is ready for the player load
+
+By default the base plugin returns a resolved plugin. If you wish the player load and play (and middlewares interception) to wait
+for some async action (i.e loading a 3rd party library) you can override and return a Promise which is resolved
+when the plugin completes all async requirements.
+
```
OrenMe, line 185:
```suggestion
Signal player that plugin has finished loading its dependencies and player can continue to loading and playing states.
Use this when your plugin requires to load 3rd party dependencies which are required for the plugin operation.
By default the base plugin returns a resolved plugin. 
If you wish the player load and play (and middlewares interception) to wait for some async action (i.e loading a 3rd party library) you can override and return a Promise which is resolved when the plugin completes all async requirements.

```

OrenMe (commented):
can we add a test here?

OrenMe (approved):

PR #374, opened 2020-11-22
feat(FEC-10669): add ability to pass options to loadMedia request

dan-ziv:

We have too many times where we want to override the provider config with data for sources object and then we need to explain to people how to do:
```
1.loadMedia()
2.configure()
3.play()
```

Instead, let's include the source config for the loadMedia request and add startTime to it (and of course not able to do preload or autoplay in this case).

New loadMedia signature will be:

```js
loadMedia(mediaInfo: ProviderMediaInfoObject, mediaOptions?: KPLoadMediaOptions)
```

### CheckLists

- [ ] Docs have been updated


src/kaltura-player.js
```diff
@@ -76,19 +94,30 @@ class KalturaPlayer extends FakeEventTarget {
...
     providerResult
       .then(
-        mediaConfig => this.setMedia(mediaConfig),
+        (providerMediaConfig: ProviderMediaConfigObject) => {
+          const mediaConfig = Utils.Object.copyDeep(providerMediaConfig);
+          if (mediaOptions) {
+            mediaConfig.playback = mediaConfig.playback || {};
+            mediaConfig.sources = mediaConfig.sources || {};
+            const {startTime} = mediaOptions;
+            if (typeof startTime === 'number') {
```
Yuvalke, line 103: @dan-ziv I think it would make a problem cause it'll be saved for the next media as we have today with startTime.
I think the better option is to set startTime on sources instead of playback config, it'll be possible to pass it on `setMedia` as well for specific media.

Yuvalke (changes requested):

src/kaltura-player.js
OrenMe: agree, this will require changes in playkit-js where startTime can come from source.
I would actually remove playkit-js startTime config from playback all together and move it to sources level and align it here in kaltura-player to merge it if it was given on playback config with deprecated warning.

OrenMe: after discussing with @dan-ziv we should handle this in separate PR.
This PR solves issue with being able to pass `mediaOptions` in run time without requiring app to do reset, configure, loadMedia and just do loadMedia.
the startTime is indeed an issue to take care of, but as mentioned it will be done separately.

OrenMe (approved):

OrenMe (approved):

src/kaltura-player.js
Yuvalke: OK, we need to add to this PR to remove the playback config from setMedia config as well.

Yuvalke (approved):

PR #370, opened 2020-11-12
feat(FEC-10656): set forceRedirectExternalStreams as true by default on IE11 and Chromecast

yairans:
### Description of the Changes

implement `getDefaultRedirectOptions` in ovp and ott player (void impl in ott)

Solves FEC-10656, [KMS-22502](https://kaltura.atlassian.net/browse/KMS-22502)

### CheckLists

- [x] changes have been done against master branch, and PR does not conflict
- [x] new unit / functional tests have been added (whenever applicable)
- [x] test are passing in local environment
- [x] Travis tests are passing (or test results are not worse than on master branch :))
- [ ] Docs have been updated

PR #381, opened 2020-11-26
fix(FEC-10729): forceRedirectExternalStreams is reset in playlist

yairans:

* Update the sources after `forceRedirectExternalStreams` is set by `loadMedia`
* Save the `player.config.sources.options` for all item unless configured for an item specifically (needed for when the app wants to config the `sources.options` for all playlist items)

Solves FEC-10729

Cont. of #370 

### CheckLists

- [ ] Docs have been updated


src/common/playlist/playlist-item.js
```diff
@@ -16,15 +16,16 @@ class PlaylistItem {
...
    * @returns {void}
    * @instance
    * @memberof PlaylistItem
    */
-  updateSources(sourcesObject: ProviderMediaConfigSourcesObject): void {
+  updateSources(sourcesObject: PKSourcesConfigObject): void {
     formats.forEach(format => {
       this._sources && (this._sources[format] = sourcesObject[format]);
     });
+    this._sources && (this._sources.options = sourcesObject.options);
```
OrenMe, line 28:
```suggestion
    if (this._sources) {
      formats.forEach(format => {
        this._sources[format] = sourcesObject[format];
      });
      this._sources.options = sourcesObject.options;
    }
```

yairans: flow requires do check inside the `forEach`

src/common/playlist/playlist-item.js
```diff
@@ -2,38 +2,41 @@
...
    * @returns {void}
    * @instance
    * @memberof PlaylistItem
    */
-  updateSources(sourcesObject: ProviderMediaConfigSourcesObject): void {
+  updateSources(sourcesObject: PKSourcesConfigObject): void {
     formats.forEach(format => {
       this._sources && (this._sources[format] = sourcesObject[format]);
     });
+    if (this._sources && sourcesObject.options) {
```
Yuvalke, line 28: doesn't it enough to replace the source object by `sourcesObject `?

src/common/playlist/playlist-manager.js
```diff
@@ -41,7 +41,7 @@ class PlaylistManager {
    */
   configure(config: ?KPPlaylistObject, entryList: ?ProviderEntryListObject) {
     if (config) {
-      this._playlist.configure(config);
+      this._playlist.configure(config, Utils.Object.getPropertyPath(this._player.config, 'sources.options'));
```
Yuvalke, line 44: doesn't it enough to send the config? `sources` already exist under config

Yuvalke (changes requested):

src/common/playlist/playlist-manager.js
Yuvalke: it's complicated that there are multiple sources from a different components

yairans: `config` it's the playlist config (id, poster etc.) it's not including the item sources

src/common/playlist/playlist-item.js
yairans: Since now I'm taking the sources from the config (not from `loadMedia` response which can be overridden by app config) you are right.

Yuvalke (approved):

PR #382, opened 2020-11-26
chore(FEC-10233): fixed async plugin loading unit tests

RoyBregman:

changes unit tests of async plugin loading:

- wrapped with try catch
- removed timed check and replaced with promise resolved / rejected
- changed dummy promises to create the promise once and not upon every getter call
- destroyed plugins and player after each test

fixes tests of FEC-10233

### CheckLists

- [ ] Docs have been updated


test/src/kaltura-player.spec.js
```diff
@@ -616,111 +616,131 @@ describe('kaltura player api', function () {
...
       PluginManager.register('asyncResolve', AsyncResolvePlugin);
       PluginManager.register('asyncReject', AsyncRejectPlugin);
     });
 
     afterEach(() => {
+      if (player._pluginManager.get('asyncResolve')) player._pluginManager.get('asyncResolve').destroy();
+      if (player._pluginManager.get('asyncReject')) player._pluginManager.get('asyncReject').destroy();
       PluginManager.unRegister('asyncResolve');
       PluginManager.unRegister('asyncReject');
+      player.destroy();
```
Yuvalke, line 629: this should make the destroy for plugins as well

Yuvalke (changes requested):

test/src/kaltura-player.spec.js
Yuvalke: So remove the destroy for plugins via plugins manager

RoyBregman: good point, done

test/src/common/plugin/test-plugins/async-resolve-plugin.js
```diff
@@ -1,16 +1,21 @@
...
 export default class AsyncResolvePlugin extends BasePlugin {
   static DELAY_ASYNC = 500;
   static isValid(): boolean {
     return true;
   }
+  promise: DeferredPromise;
 
+  constructor(name: string, player: Object, config: Object) {
+    super(name, player, config);
+    this.promise = Utils.Object.defer();
```
Yuvalke, line 13: I think it preferred to move it to ready to make sure the timeout will resolve after your check of promise fulfill order happened correctly.

RoyBregman: Actually i liked this better as it is more like a real life example.
Putting the setTimeout in the getter will make need to add some flagging mechanism and also if I define two and more plugins then the time for them to be ready will the accumulation of their setTimeouts instead of the max setTimeout like in real life (and also how I implemented it now)

yairans (approved):

PR #383, opened 2020-11-26
fix(FEC-10776): set the plugins event registration after kaltura player internal events

Yuvalke:

Issue: plugins register before UI/provider and internal handler in Kaltura player which could cause incorrect behavior in the player after plugins handler.
Solution: configure the plugins after internal logic - seems we would like to handled last.
UI components that come from plugins adding by addComponent API so external UI components of Kaltura player will update via config.
For example - when the plugin registers on the contractor to error and makes a reset it'll handle the event before UI/provider and Kaltura player handle it and cause the error overlay to appear on UI, plugin call to reset didn't clear this state after it changed, it cleared it before UI changed it and does nothing.

### CheckLists

- [ ] Docs have been updated


OrenMe (approved):
LGTM.
@yairans do you see any risks with plugins and UI?

src/kaltura-player.js
```diff
@@ -75,6 +74,7 @@ class KalturaPlayer extends FakeEventTarget {
     Object.values(CoreEventType).forEach(coreEvent => this._eventManager.listen(this._localPlayer, coreEvent, e => this.dispatchEvent(e)));
     this._addBindings();
     this._playlistManager.configure(options.playlist);
+    this.configure({plugins});
```
yairans, line 77: The `UIWrapper` uses the `player.uiComponents` which filled by the plugins while their initialization. So we can't move it after the `UIWrapper` initialization.

yairans (changes requested):

src/kaltura-player.js
OrenMe: So the correct flow here is to actually fix the UI manager issue.
@Yuvalke the correct expectation is that if someone calls reset on player it will actually reset the UI as well, even if it is in transit.
Issue is that as we call reset on error event and event didn't yet reached UI we can't really stop it - chicken and egg kind of situation.
I suggest you and @yairans discuss this so @yairans can have more context and let's try to figure it out.

Yuvalke: @yairans @OrenMe we use pass on configure with setConfig and pass the uiComponents there instead, I'll need to verify it with plugins like this

OrenMe: @Yuvalke plugins expose UI components, it is not just app so need them to be available before.
But now with dynamic injection this is not true any more and I assume we can find a solution where UI is configured and then plugins and then we register plugins UI components on UI.
@yairans FYI

Yuvalke: https://github.com/kaltura/playkit-js-ui/pull/563 - implement the add UI component which provides an option to fix it

Yuvalke: @yairans updated as we discuss, external UI components will update via config and plugins will update via API.

src/common/ui-wrapper.js
```diff
@@ -16,7 +16,6 @@ class UIWrapper {
 
   constructor(player: KalturaPlayer, options: KPOptionsObject) {
     const config: KPUIOptionsObject = options.ui;
-    config.uiComponents = [...(player.uiComponents || []), ...(config.uiComponents || [])];
```
OrenMe, line 19: it is possible for app to add components as well so can't remove this

OrenMe (changes requested):

src/common/ui-wrapper.js
Yuvalke: why not? it just merge the plugins uiComponents with config uiComponents, it shouldn't change and then uiComponents will pass as is

OrenMe: need to make sure that if app passes the uiComponents config it is passed.
We only discussed the plugins which is internal, but an app can do:
```javascript
KalturaPlayer.setup({
  plugins: {
    uiComponents: []
  }
})
```

Yuvalke: Yes, sure. Why not? Nobody remove it, it pass on config directly to ui manager as it was before.
Those lines merging the uiComponents from plugins and config.uiComponents which doesn't need it anymore its done by addComponent API instead

yairans:
```
KalturaPlayer.setup({
  ui: {
    uiComponents: []
  }
})
```
@OrenMe

OrenMe: correct :-)

Yuvalke: so we're good, right?

OrenMe: No, app provided config will stop working

OrenMe: Got ya, you are correct, the config is still passing 👍

yairans: So `player.uiComponents` is useless. maybe we can remove it (don't think anyone uses it)

Yuvalke: Didn't remove it since it's API

OrenMe: I think we only used it when we passed the uiComponents from playkit when plugins were in it.
https://github.com/kaltura/kaltura-player-js/pull/264
It is safe to remove as we alway supported adding it only on initial config until we recently added the dynamic injection API which has it's own code paths.

Yuvalke: I mean if customers use it as getter, anyway if it's ok by you I'll remove it

OrenMe: @esakal can you please review the removal of `uiComponents` getter.
It seems this is only used for setting it but maybe `contrib` is using it?

yairans (approved):

OrenMe (approved):

PR #393, opened 2020-12-22
fix(FEC-10797): back-end bumper config is left from previous media played

yairans:

Reset the provider plugins on reset but keep it on the playlist item for the next time the item will be played (use `setMedia`)

Solves FEC-10797

### CheckLists

- [ ] Docs have been updated


src/kaltura-player.js
```diff
@@ -772,18 +774,25 @@ class KalturaPlayer extends FakeEventTarget {
     this._localPlayer.detachMediaSource();
   }
 
-  _mergeProviderPluginsConfig(providerPluginsConfig: KPPluginsConfigObject): KPPluginsConfigObject {
+  mergeProviderPluginsConfig(providerPluginsConfig: KPPluginsConfigObject): KPPluginsConfigObject {
```
OrenMe, line 777: don;t think we would want to make this a kaltura player API(removing the `_`)

yairans: it's called from the playlist manager

OrenMe: Still, not a valid API on the player, it is also externalized.

OrenMe: if this becomes a helper function move it to a seperate utils and use it without side-effects.
As it is private you can change API to return the object with both providers config as before and also the cached app config and use the one you need where you need it.
Exposing this on the API is problematic as it will never be used and will just raise questions.

OrenMe (approved):

OrenMe (approved):

OrenMe (approved):

PR #405, opened 2021-01-31
fix(FEC-10872): loadMedia returns the provider response instead of the updated one

dan-ziv:

Since FEC-10669 loadMedia gets options which could change the media returned by the provider, but loadMedia still returns a promise holds the provider's response.
Solution: resolve the provider promise inside the callback with returning the updated object.

### CheckLists

- [ ] Docs have been updated


Yuvalke (dismissed):

src/kaltura-player.js
```diff
@@ -119,28 +119,35 @@ class KalturaPlayer extends FakeEventTarget {
...
+            this._appPluginConfig = mergedPluginsConfigAndFromApp[1];
+            this.configure(getDefaultRedirectOptions(this.config, mediaConfig));
+            this.setMedia(mediaConfig);
+            return mediaConfig;
+          },
+          e => {
+            this._localPlayer.dispatchEvent(
+              new FakeEvent(CoreEventType.ERROR, new Error(Error.Severity.CRITICAL, Error.Category.PLAYER, Error.Code.LOAD_FAILED, e))
+            );
+            reject(e);
```
OrenMe, line 143: what did we use to return as an error before?

dan-ziv: We returned the `providerPromise`, so if it was rejected, you returned rejected promise.

OrenMe: and what was the object of it?

OrenMe: I think we should keep the isolation we have today and return the error object with the error nested in it

OrenMe: WDYT?

dan-ziv: > and what was the object of it?

undefined

> I think we should keep the isolation we have today and return the error object with the error nested in it

Good point, fixed.

OrenMe (dismissed):

Yuvalke (dismissed):

src/kaltura-player.js
```diff
@@ -119,28 +119,34 @@ class KalturaPlayer extends FakeEventTarget {
...
     this._uiWrapper.setLoadingSpinnerState(true);
-    const providerResult = this._provider.getMediaConfig(mediaInfo);
-    providerResult
-      .then(
-        (providerMediaConfig: ProviderMediaConfigObject) => {
-          const mediaConfig = Utils.Object.copyDeep(providerMediaConfig);
-          if (mediaOptions) {
-            mediaConfig.sources = mediaConfig.sources || {};
-            mediaConfig.sources = Utils.Object.mergeDeep(mediaConfig.sources, mediaOptions);
+    return new Promise((resolve, reject) => {
```
yairans, line 122: @dan-ziv Please change https://github.com/kaltura/kaltura-player-js/blob/master/src/common/playlist/playlist-manager.js#L286 to use the returning value instead of the config (sources and plugins)

yairans (changes requested):

src/kaltura-player.js
dan-ziv: I didn't changed anything in the return value, just the way we returning it.

yairans: what you returning now is containing the merge plugins and the redirect options so no need to take them from the config in the playlist manager

yairans (changes requested):

test/src/kaltura-player.spec.js
```diff
@@ -89,8 +89,8 @@ describe('kaltura player api', function () {
           errorEventTriggered = true;
         });
         kalturaPlayer.loadMedia({}).catch(error => {
-          error.data.should.equal('Missing mandatory parameter');
-          error.success.should.be.false;
+          error.data.data.should.equal('Missing mandatory parameter');
```
yairans, line 92: why this being changed? `data.data` seems weird also this is breaking change of the loadMedia catch listeners

yairans (changes requested):

yairans (approved):

PR #408, opened 2021-02-02
feat(FEC-10041): playAdsWithMSE with DAI detach the playback and ad

Yuvalke:

Issue: ad and playback have the same content URI.
Solution: check if ads are imadai and exclude the logic for attaching detach.

### CheckLists

- [ ] Docs have been updated


OrenMe (commented):
Maybe do it more generic, the plugin decorator can expose data if requires detach/attach and then we ask to get current active decorator and ask it if it needs detach/attach.
Base will always return true and plugins implementing decorator can override and return false

Yuvalke:
@OrenMe I thought about that but why wouldn't we pass it on adLoaded params? sounds a bit less robust I think

OrenMe:
@Yuvalke my point here is to make it more generic and not check in kaltura-player for specific plugin names. If it makes more sense to pass it in another manner then raise it and we can consider.

yairans (dismissed):

PR #413, opened 2021-02-10
feat(FEC-10640): add api to get the playlist current working item index

yairans:

add `current` api to the playlist manager and add `index` field to the playlist item
### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/playlist/playlist-manager.js
+++ b/src/common/playlist/playlist-manager.js
@@ -95,8 +95,8 @@ class PlaylistManager {
   playNext(): void {
     PlaylistManager._logger.debug('playNext');
     const next = this._playlist.getNext(true);
-    if (next.item) {
-      this._setItem(next.item, next.index);
+    if (next) {
+      this._setItem(next);
     }
   }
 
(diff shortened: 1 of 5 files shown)


OrenMe (dismissed):

OrenMe:
this refactor actually makes the code nicer :-)

src/common/playlist/playlist-manager.js
```diff
@@ -95,8 +95,8 @@ class PlaylistManager {
   playNext(): void {
     PlaylistManager._logger.debug('playNext');
     const next = this._playlist.getNext(true);
```
Yuvalke, line 97: It's not about the change but maybe you will know why doesn't it use the getter next and force to get the next one even the loop config is false?

Yuvalke (dismissed):

src/common/playlist/playlist-manager.js
yairans: loop it's for the UI and for automatically next playing only. for the api loop it's always true

yairans: loop it's for the UI and for automatically next playing only. for the api loop it's always true

PR #415, opened 2021-02-16
feat(FEC-10768): expose in-stream DASH thumbnails

dan-ziv:

* Move all thumbnail related logic to `ThumbnailManager`.
* Added new API `getThumbnail(time?: number): ?ThumbnailInfo`.

### CheckLists

- [ ] Docs have been updated


src/kaltura-player.js
```diff
@@ -452,6 +453,21 @@ class KalturaPlayer extends FakeEventTarget {
...
+    if (this._thumbnailManager) {
+      if (!time) {
+        // If time isn't supplied, return thumbnail for player's current time
+        if (!isNaN(this.currentTime)) {
+          time = this.currentTime;
+        } else {
+          return null;
+        }
+      }
+      time = this.isLive() ? time + this.getStartTimeOfDvrWindow() : time;
```
dan-ziv, line 466: @OrenMe this is an example where we need the live absolute time (the video element time) and to separate the APIs.
For now I did the transformation here, but I'm still debating if it should be on the UI/application side.
Please give your thoughts.

src/common/thumbnail-manager.js
```diff
@@ -21,7 +21,7 @@ class ThumbnailManager {
   }
 
   getThumbnail(time: number): ?ThumbnailInfo {
-    if (this._thumbnailConfig) {
+    if (this._isValidThumbnailConfig()) {
```
Yuvalke, line 24: @dan-ziv didn't you create an ImageTrack for Kaltura thumbnail as well?

src/common/thumbnail-manager.js
```diff
@@ -0,0 +1,84 @@
+// @flow
+import {Utils, ThumbnailInfo, MediaType} from '@playkit-js/playkit-js';
+import evaluate from './utils/evaluate';
+
+const DefaultThumbnailConfig: Object = {
+  thumbsWidth: 164,
```
Yuvalke, line 6: what are that numbers? 164,92,100? there are multiple options? I also see height doesn't exist in the previous implementation. can't you get a list of thumbnails?

dan-ziv: Those are the default values we are using with the thumbnail service (see [here](https://developer.kaltura.com/api-docs/Engage_and_Publish/kaltura-thumbnail-api.html)).
The height (92) is also the default value we are using, but I added also its definition because I need it as part of the calculation when returning `ThumbnailInfo`.

src/common/thumbnail-manager.js
dan-ziv: Currently no. I've thoughts about it, but I don't no if it's right - tracks are something that defined and comes from the manifest and this is not the case here. So it will maybe looks weird and buggy to see a track comes out of no where.
BTW can you think how to add track from outside of playkit? probably this is something that if we will do should be pass through the config...

Yuvalke: We already have something like this, we have external tracks which comes from config
BTW don't we get a list for thumbnail images?

dan-ziv: Yes, but in external tracks the config explicitly defines it - not like here when its happens "magically" since this data comes from the backend (usually, except some OTT partners).
Regarding your Q, I think you are confusing with posters retrieved from OTT provider, but in this case this is not URLs of Kaltura's thumbnail service so the `thumbsSprite` result will be an empty string (this is why the regex check).

src/common/ui-wrapper.js
```diff
@@ -91,10 +90,9 @@ class UIWrapper {
     this.setConfig({hasError: false}, 'engine');
   }
 
-  setSeekbarConfig(mediaConfig: KPMediaConfig, uiConfig: KPUIOptionsObject): void {
+  setSeekbarConfig(uiConfig: KPUIOptionsObject, thumbnailConfig: ?SeekbarConfig): void {
```
yairans, line 93: this is api (via `kalturaPlayer.ui.setSeekbarConfig`) i'm not sure we can break it

src/common/thumbnail-manager.js
```diff
@@ -0,0 +1,84 @@
...
+      thumbnailInfo.y = thumbsHeight;
+      thumbnailInfo.url = thumbsSprite;
+      thumbnailInfo.height = thumbsHeight;
+      thumbnailInfo.width = thumbsWidth;
+      return thumbnailInfo;
+    }
+  };
+
+  _buildKalturaThumbnailConfig = (mediaConfig: KPMediaConfig): ?SeekbarConfig => {
+    const seekbarConfig = Utils.Object.getPropertyPath(this._player.config.ui, 'components.seekbar');
```
yairans, line 53: we are about to remove the ui config from the core player so this should be passed from kaltura player

src/common/thumbnail-manager.js
```diff
@@ -0,0 +1,84 @@
...
+
+  getThumbnail(time: number): ?ThumbnailInfo {
+    if (this._isUsingKalturaThumbnail()) {
+      return this._convertKalturaThumbnailToThumbnailInfo(time);
+    }
+    return this._player.getThumbnail(time);
+  }
+
+  getKalturaThumbnailConfig(): ?SeekbarConfig {
+    return this._thumbnailConfig;
```
yairans, line 31: this is only for the ui until we change it to support the new api right?

yairans (changes requested):

src/common/ui-wrapper.js
dan-ziv: It shouldn't be an API and you should have hidden it once you exposed the `ui` namespace. If someone wants to set seekbar config it should be done via player config - this method is internal for us to communicate with the ui manager.
Anyway it will be removed and will be useless once the UI will work with the `getThumbnail` API.

src/common/thumbnail-manager.js
dan-ziv: yes.

src/common/thumbnail-manager.js
Yuvalke: I think I'm missing something cause I didn't find those default values, also I see you have an option to request different parameters from the thumbnail so I wonder why it's not open for user selection?

src/common/thumbnail-manager.js
```diff
@@ -0,0 +1,84 @@
...
+const THUMBNAIL_REGEX = /.*\/p\/\d+\/(?:[a-zA-Z]+\/\d+\/)*thumbnail\/entry_id\/\w+\/.*\d+/;
+const THUMBNAIL_SERVICE_TEMPLATE: string = '{{thumbnailUrl}}/width/{{thumbsWidth}}/height/{{thumbsHeight}}/vid_slices/{{thumbsSlices}}/ks/{{ks}}';
+
+class ThumbnailManager {
+  _player: Player;
+  _thumbnailConfig: ?SeekbarConfig;
+
+  constructor(mediaConfig: KPMediaConfig, player: Player) {
+    this._player = player;
+    this._thumbnailConfig = this._buildKalturaThumbnailConfig(mediaConfig);
```
Yuvalke, line 20: Maybe I don't remember but why wouldn't we create track? I saw that you actually can request from thumbnail service multiple different parameters so you can provide parameters that will change the track on every get thumbnail request with different parameters and then it reasonable(same we do for external track) not sure if it's needed to expose all the options but I think we take it into considartion.

src/common/thumbnail-manager.js
dan-ziv: fixed

yairans (dismissed):

src/common/thumbnail-manager.js
dan-ziv: You can see the defaults values currently [here](https://github.com/kaltura/kaltura-player-js/blob/master/src/common/utils/thumbs.js#L5) and [here](https://github.com/kaltura/kaltura-player-js/blob/master/src/common/utils/thumbs.js#L6).
But I think you're a bit deviated from the PR subject. This is not a discussion about how the thumbnail service is implemented - this is the way we supported it from player's day one. I've just copied the kaltura thumbnail logic to a new file. If you want to open it to discussion let's setup a team meeting.

PR #417, opened 2021-02-22
fix(FEC-10281): chromecast does not work after playing it once and trying it on another video

dan-ziv:

Call also to remote player destroy on player destroy.

### CheckLists

- [ ] Docs have been updated


src/common/cast/remote-player-manager.js
```diff
@@ -47,6 +47,13 @@ class RemotePlayerManager {
     return false;
   }
 
+  static destroy(type?: string): void {
+    const remotePlayer = RemotePlayerManager._getRemotePlayer(type);
```
Yuvalke, line 51: shouldn't it remove all the remote players?

src/kaltura-player.js
```diff
@@ -326,6 +326,7 @@ class KalturaPlayer extends FakeEventTarget {
     this._localPlayer.destroy();
     this._eventManager.destroy();
     this._viewabilityManager.destroy();
+    RemotePlayerManager.destroy();
```
Yuvalke, line 329: it won't destroy the remote player without type, so I guess it should remove them all on destroy as I wrote in the previous comment, am I right?

Yuvalke (changes requested):

src/kaltura-player.js
dan-ziv: Good point. FIXED

yairans (approved):

PR #418, opened 2021-02-28
fix(FEC-11037): multiple decorator exist after destroy plugin with decorator

Yuvalke:

Issue: pass the entire plugin instead of a specific function to avoid memory leak/non-relevant instance.
Solution: pass the only function that relevant and his name to verify the same plugin won't register twice

### CheckLists

- [ ] Docs have been updated


src/kaltura-player.js
```diff
@@ -809,7 +808,7 @@ class KalturaPlayer extends FakeEventTarget {
             }
 
             if (typeof plugin.getEngineDecorator === 'function') {
-              registerEngineDecoratorProvider(plugin);
+              this._localPlayer.registerEngineDecoratorProvider(name, {getEngineDecorator: plugin.getEngineDecorator.bind(plugin)});
```
dan-ziv, line 811: Seems weird to write it that way.
Why not just pass the handler? and if you really need `getEngineDecorator` API why not to define `EngineDecoratorProvider` class which accepts the plugin and expose all needed fields ?
Something like:
```js
this._localPlayer.registerEngineDecoratorProvider(new EngineDecoratorProvider(plugin));

class EngineDecoratorProvider {
  constructor(plugin) {
    this.name = plugin.getName();
    this.hander = plugin.getEngineDecorator.bind(plugin);
  }

  getEngineDecorator() {
    return this.hander;
  }

  getName() {
    return this.name;
  }
}

```

src/engine-decorator-provider.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+export default class EngineDecoratorProvider implements IEngineDecoratorProvider {
```
dan-ziv, line 2: Great @Yuvalke, but few things are missing (I just gave pseudo code example):
1. Add type annotations to methods and members (also need to mark members as private `_`).
2. Don't export default but named (caused us some issues in the past).
3. Export `EngineDecoratorProvider` from playkit (this where the engine decorator framework is defined).

dan-ziv (approved):

OrenMe (approved):

PR #423, opened 2021-03-08
feat(FEC-10941): Use In-Stream DASH thumbnails on the timeline

dan-ziv:

Remove setter for UI seekbar config since the UI is moving to work with the thumbnail API.

### CheckLists

- [ ] Docs have been updated


src/common/thumbnail-manager.js
```diff
@@ -37,7 +37,8 @@ class ThumbnailManager {
 
   _convertKalturaThumbnailToThumbnailInfo = (time: number): ?ThumbnailInfo => {
     if (this._thumbnailConfig) {
-      const {thumbsSprite, thumbsWidth, thumbsHeight, thumbsSlices} = this._thumbnailConfig;
+      const {thumbsSprite, thumbsWidth, thumbsSlices} = this._thumbnailConfig;
+      const {thumbsHeight} = DefaultThumbnailConfig;
```
yairans, line 41: so the height is not configurable anymore ? why?

yairans (changes requested):

src/common/thumbnail-manager.js
dan-ziv: As per our patch with the backend cache issue, we are not adding the height to the thumbnail service request, so no sense of make it configurable.

yairans: you mean the height is fixed when we getting the thumbnail from kaltura? 
this is because kaltura always returns one line sprite and not matrix? and the thumb height is the sprite height?
so maybe better to remove height from here at all and ThumbnailInfo should return the sprite height for this 
I just wonder if kaltura BE will change the default height we get incorrect value here

dan-ziv: The `ThumbnailInfo` shouldn't return the sprite height since thumbnail info returns info regarding specific thumbnail. As a consumer of this API you just want to know what are the thumb's dimensions and coordinates.
Thumb height is something that included in the data of the `ImageTrack`. In this case of Kaltura thumbnail, we are still not creating image track. 
As for the BE, what you are saying was true even before this change, as we used hard-coded height and width in the UI. This is why in the original implementation I was added the call with explicit values.

yairans: i meant when the sprite is one line the sprite height is actually is the thumb height so why not to use it?

dan-ziv: How do u know what is the actual height?

yairans (approved):

PR #425, opened 2021-03-08
fix(FEC-11062): ad layout doesn't work when IMA DAI configured

Yuvalke:

Issue: play ad now to use the active plugin with play ad now API.
Solution: instead of empty implementation on ad plugins, implement it only where it's needed, and check if there's plugin with this API available to play the ad.

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/controllers/ads-controller.js
+++ b/src/common/controllers/ads-controller.js
@@ -280,7 +280,7 @@ class AdsController extends FakeEventTarget implements IAdsController {
   }
 
   _playAdBreak(adBreak: RunTimeAdBreakObject): void {
-    const adController = this._adsPluginControllers.find(controller => !this._isBumper(controller));
+    const adController = this._adsPluginControllers.find(controller => !this._isBumper(controller) && typeof controller.playAdNow === 'function');
     if (adController) {
       adBreak.played = true;
       this._adIsLoading = true;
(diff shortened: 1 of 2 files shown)


src/common/controllers/ads-controller.js
```diff
@@ -280,8 +280,8 @@ class AdsController extends FakeEventTarget implements IAdsController {
   }
 
   _playAdBreak(adBreak: RunTimeAdBreakObject): void {
-    const adController = this._adsPluginControllers.find(controller => !this._isBumper(controller));
-    if (adController) {
+    const adController = this._adsPluginControllers.find(controller => !this._isBumper(controller) && typeof controller.playAdNow === 'function');
+    if (adController && typeof adController.playAdNow === 'function') {
```
yairans, line 284: why is needed? for flow?

src/common/controllers/ads-controller.js
```diff
@@ -280,8 +280,8 @@ class AdsController extends FakeEventTarget implements IAdsController {
   }
 
   _playAdBreak(adBreak: RunTimeAdBreakObject): void {
-    const adController = this._adsPluginControllers.find(controller => !this._isBumper(controller));
-    if (adController) {
+    const adController = this._adsPluginControllers.find(controller => !this._isBumper(controller) && typeof controller.playAdNow === 'function');
```
yairans, line 283: you can remove `!this._isBumper` check

yairans (changes requested):

src/common/controllers/ads-controller.js
Yuvalke: Yes

yairans (approved):

PR #452, opened 2021-05-30
fix(FEC-10994): Invalid version reported by players

Yuvalke:

Issue: some clients load the JS script with one uiconf but use another one in the call to setup or they don't provide it at all.
Solution: always take the uiconf config from the bundle object.

related changes:
https://github.com/kaltura/server/pull/10800

### CheckLists

- [ ] Docs have been updated


src/common/utils/setup-helpers.js
```diff
@@ -292,7 +275,7 @@ function getDefaultOptions(options: PartialKPOptionsObject): KPOptionsObject {
   };
   Utils.Object.mergeDeep(defaultOptions, options);
   if (defaultOptions.provider.uiConfId) {
```
yairans, line 277: we want to support also when uiconf id no passed from the app. so can remove this if

yairans (changes requested):

yairans (approved):

yairans:
@Yuvalke please add the server change url in the desc

OrenMe (approved):

PR #478, opened 2021-08-12
fix(FEC-11441): Default kava details are reported with wrong (unknown) player version

yairans:

add the product version to the default kava impression and to the client tag param
extract `validateProviderConfig` from `validateConfig` to call it after getting the uiconf config

### CheckLists

- [ ] Docs have been updated


flow-typed/types/kaltura-player-options.js
```diff
@@ -34,7 +34,8 @@ declare type PartialKPOptionsObject = {
   playlist?: KPPlaylistObject,
   dimensions?: PKDimensionsConfig,
   ui?: KPUIOptionsObject,
-  cast?: {[key: string]: any}
+  cast?: {[key: string]: any},
+  productVersion?: string
```
OrenMe, line 38: this value shouldn't be configurable, please remove from config object

OrenMe: on second thought, we pass it in the config object? we need to extract it to be internal cause it can be easily changed as a config.

yairans: do you mean to delete it from the config?

PR #488, opened 2021-09-06
feat(FEC-11540): CuePoint Manager

semarche-kaltura:

Added cue-point-manager that has public API to add and manage text-tracks
Cover cue-point-manager by tests

### CheckLists

- [ ] Docs have been updated


yairans:
@semarche-kaltura please clone this ticket to FEC and change the label

src/common/cuepoint-manager.js
```diff
@@ -0,0 +1,72 @@
...
+  startTime: number;
+  endTime: number;
+}
+
+export class CuePointManager {
+  Event = {
+    TIMED_METADATA_ADDED: 'timedmetadataadded'
+  };
+  Type = {
+    CuePointsTextTrack: 'CuePoints',
```
yairans, line 15: enum keys are upper case - CUE_POINTS_TEXT_TRACK

src/common/cuepoint-manager.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+  _player: KalturaPlayer;
+  _textTrack: TextTrack | null = null;
+
+  constructor(player: KalturaPlayer) {
+    this._player = player;
+  }
+
+  _addTextTrack = () => {
+    this._textTrack = this._player.addTextTrack('metadata', this.Type.CuePointsTextTrack);
```
yairans, line 26: use TextTrack.KIND.METADATA

src/kaltura-player.js
```diff
@@ -90,6 +92,7 @@ class KalturaPlayer extends FakeEventTarget {
     this._viewabilityManager = new ViewabilityManager(this.config.viewability);
     this._uiWrapper = new UIWrapper(this, Utils.Object.mergeDeep(options, {ui: {logger: {getLogger, LogLevel}}}));
     this._serviceProvider = new ServiceProvider(this);
+    this._cuepointManager = new CuePointManager(this);
```
yairans, line 95: add reset and destroy handlers

src/common/cuepoint-manager.js
yairans: I think its better to separate the enum config and expose in from kaltura player index js (like playlist)

src/common/cuepoint-manager.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+  Type = {
+    CuePointsTextTrack: 'CuePoints',
+    CuePointKey: 'CuePoint'
+  };
+  _player: KalturaPlayer;
+  _textTrack: TextTrack | null = null;
+
+  constructor(player: KalturaPlayer) {
+    this._player = player;
```
OrenMe, line 22: on reset need to clear cuepoints, they are media session dependent, so cuepoints of one media are not relevant for next media.
Listen to reset on kaltura player and call `clearAllCuePoints()`

On destroy need to do the same.

OrenMe (changes requested):

src/common/cuepoint/cuepoint-event-type.js
```diff
@@ -0,0 +1,22 @@
...
+ *   console.log(e.payload);
+ * };
+ */
+const CuePointEventType: {[event: string]: string} = {
+  /**
+   * Fires when new cue-points added to text-track.
+   * @event TIMED_METADATA_ADDED
+   * @memberof CuePointEventType
+   */
+  TIMED_METADATA_ADDED: `${namespace}-timedmetadataadded`
```
yairans, line 19: maybe add here also the `TIMED_METADATA`, (the fact it comes from the player it's internal detail, it will be clearer if both events are coming from cue point). but if so need to remove the `kaltura-player` namespace

src/common/cuepoint/cuepoint-manager.js
```diff
@@ -0,0 +1,87 @@
...
+
+export class CuePointManager {
+  _player: KalturaPlayer;
+  _eventManager: EventManager;
+  _textTrack: TextTrack | null = null;
+
+  constructor(player: KalturaPlayer) {
+    this._player = player;
+    this._eventManager = new EventManager();
+    this._eventManager.listen(player, player.Event.Core.PLAYER_RESET, () => this.reset());
```
yairans, line 20: not need. call to CuePointManager.reset on kaltura-player reset. like https://github.com/kaltura/kaltura-player-js/blob/master/src/kaltura-player.js#L320

src/common/cuepoint/cuepoint-event-type.js
```diff
@@ -0,0 +1,26 @@
...
+   * @event TIMED_METADATA_ADDED
+   * @memberof CuePointEventType
+   */
+  TIMED_METADATA_ADDED: 'timedmetadataadded',
+  /**
+   * Fired when the timed metadata triggered
+   * @event TIMED_METADATA
+   * @memberof CuePointEventType
+   */
+  TIMED_METADATA: 'timedmetadata'
```
OrenMe, line 23: up until now `timedmetadata` was not for cuepoints and was part of core events. we can also add `timedmetadataadded` to it, no need to create additional nested/namespaced event object

src/common/cuepoint/cuepoint-manager.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+  };
+
+  reset = () => {
+    this.clearAllCuePoints();
+  };
+
+  destroy = () => {
+    this.reset();
```
OrenMe, line 81: maybe I missed it but no one is calling destroy, did I miss anything?

src/common/cuepoint/cuepoint-manager.js
```diff
@@ -0,0 +1,83 @@
...
+import {CUE_POINTS_TEXT_TRACK, CUE_POINT_KEY} from './cuepoint-type';
+import {CuePointEventType} from './cuepoint-event-type';
+
+interface CuePoint {
+  id: string;
+  startTime: number;
+  endTime: number;
+}
+
+export class CuePointManager {
```
OrenMe, line 12: @semarche-kaltura any reason for defining all methods as arrow function? this is a class and it is instantiated so `this` is always the correct context

OrenMe (changes requested):

src/common/cuepoint/cuepoint-event-type.js
semarche-kaltura: I created nested/namespaced event object because of PR comments that I received to the PR

src/common/cuepoint/cuepoint-manager.js
semarche-kaltura: seems like we don't need `destroy` method

src/common/cuepoint/cuepoint-manager.js
semarche-kaltura: changed

src/common/cuepoint/cuepoint-manager.js
OrenMe: why not? kaltura player destroy doesn't call its reset method so it won't get cleared. Although maybe it is not a concern but better hook it if it is there so in future other team members won't get confused.

src/common/cuepoint/cuepoint-event-type.js
OrenMe: ok, so @yairans or @RoyBregman - `timedmetadata` is part of core already, so we don't need to move it and actually can't/shouldn't.
I suggest not to create a namespace for this only for 1 event and move it to the core library, not first time we ad an event in core for benefit of external use, when the bigger context makes more sense then defining it outside

src/common/cuepoint/cuepoint-manager.js
semarche-kaltura: ok

src/common/cuepoint/cuepoint-event-type.js
semarche-kaltura: done

yairans (dismissed):

yairans (dismissed):

OrenMe (dismissed):

yairans (dismissed):

yairans (approved):

PR #495, opened 2021-09-20
fix(FEC-11156): cast button appears for one player only when multiple players configured on the page and casting failed

SivanA-Kaltura:

Use RemotePlayerManager instance instead of static to support multiple players
Add API to mark player where the cast button was clicked
Fixes FEC-11156

Related PRs:
https://github.com/kaltura/playkit-js-cast-sender/pull/70
https://github.com/kaltura/playkit-js-ui/pull/640

### CheckLists

- [ ] Docs have been updated


src/common/cast/remote-player-manager.js
```diff
@@ -21,39 +24,49 @@ class RemotePlayerManager {
     }
   }
 
-  static load(castConfig: Object, remoteControl: RemoteControl): void {
+  load(castConfig: Object, player: KalturaPlayer): void {
+    ++RemotePlayerManager._loadedPlayerCount;
```
SivanA-Kaltura, line 28: use count per type

src/common/cast/base-remote-player.js
```diff
@@ -309,6 +310,10 @@ class BaseRemotePlayer extends FakeEventTarget implements IRemotePlayer {
     return true;
   }
 
+  isCastInitiator(): boolean {
+    return this._isCastInitiator;
```
yairans, line 314:
```suggestion
    return this._isCastInitiator || RemotePlayerManager._loadedPlayerCount === 1;
```
so you can remove the RemotePlayerManager.load

src/common/cast/remote-player-manager.js
```diff
@@ -21,39 +24,49 @@ class RemotePlayerManager {
...
     const registry = RemotePlayerManager._registry;
     registry.forEach((RemotePlayer: Function, type: string) => {
       RemotePlayerManager._logger.debug(`Load remote player of type ${type}`);
-      RemotePlayerManager._remotePlayers.set(type, new RemotePlayer(castConfig, remoteControl));
+      const remotePlayer = new RemotePlayer(castConfig, new RemoteControl(player), player.config.targetId);
+      remotePlayer.isCastInitiator = () => remotePlayer._isCastInitiator || RemotePlayerManager._loadedPlayerCount === 1;
```
yairans, line 33: see my comment above

src/common/cast/remote-player-manager.js
yairans: what if a player destroyed, need to decrease the counter no?

SivanA-Kaltura: right now the remote players are only destroyed all at once and not one at a time

src/common/cast/base-remote-player.js
SivanA-Kaltura: it's a counter by type and we don't have a type in the base class.
also I don't like putting a specific logic in the base class in case the children want to override it, which is why i put the generic logic of returning _isCastInitiator.

src/common/cast/remote-player-manager.js
SivanA-Kaltura: so the question is whether one kaltura player on the page might be destroyed while other kaltura players are not destroyed

src/common/cast/base-remote-player.js
yairans: 1. if the child has the type the base has it also (this refers to the child)
2. if a child wants to override it's still possible, just implement `isCastInitiator` (like you do for all)

src/common/cast/remote-player-manager.js
yairans: sure. the app can destroy a specific instance.

src/common/cast/base-remote-player.js
SivanA-Kaltura: I don't like having this logic in the parent, I'll just implement it in the sender instead

src/kaltura-player.js
```diff
@@ -442,11 +443,15 @@ class KalturaPlayer extends FakeEventTarget {
   }
 
   startCasting(type?: string): Promise<*> {
-    return RemotePlayerManager.startCasting(type);
+    return this._remotePlayerManager.startCasting(type);
+  }
+
+  setIsCastInitiator(isCastInitiator: boolean) {
+    this._remotePlayerManager.setIsCastInitiator(undefined, isCastInitiator);
```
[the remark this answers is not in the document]
SivanA-Kaltura: it should actually be "chromecast"

src/common/cast/remote-player-manager.js
```diff
@@ -21,39 +23,47 @@ class RemotePlayerManager {
...
+    const remotePlayers = this._remotePlayers;
     Array.from(remotePlayers.values()).forEach(remotePlayer => remotePlayer.destroy());
   }
 
-  static _getRemotePlayer(type?: string): ?Object {
-    const remotePlayers = RemotePlayerManager._remotePlayers;
+  setIsCastInitiator(type?: string, isCastInitiator: boolean) {
+    const remotePlayer = this._getRemotePlayer(type);
+    if (remotePlayer) {
+      remotePlayer._isCastInitiator = isCastInitiator;
```
yairans, line 61: it's private. expose setter

src/kaltura-player.js
```diff
@@ -78,6 +79,8 @@ class KalturaPlayer extends FakeEventTarget {
    */
   _autoPaused: boolean = false;
 
+  RemotePlayerType: Object = RemotePlayerType;
```
yairans, line 82: not needed. use import like:
```
import {cast} from 'kaltura-player-js;
const {RemotePlayerType} =  cast;
```

SivanA-Kaltura: how do I access it from ui then ?

SivanA-Kaltura: I need to expose it so ui can access it though

yairans: like the code I wrote

yairans: already exposed 
https://github.com/kaltura/kaltura-player-js/blob/master/src/index.js#L51
https://github.com/kaltura/kaltura-player-js/blob/master/src/common/cast/index.js#L25

yairans (approved):

PR #510, opened 2021-12-23
fix(FEC-11792): slides does not appear in dual video

yairans:

fix `getCuePointById` to return the found cue point
use `some` instead of `forEackh` to break the loop once a cue point found by id
protect the `removeCuePoint`

Solves FEC-11792

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/cuepoint/cuepoint-manager.js
+++ b/src/common/cuepoint/cuepoint-manager.js
@@ -51,7 +51,7 @@ export class CuePointManager {
       if (cuePoint) {
         cuePointById = cuePoint;
       }
-      return !cuePointById;
+      return !cuePoint;
     });
     return cuePointById;
   }
--- a/src/common/cuepoint/cuepoint-manager.js
+++ b/src/common/cuepoint/cuepoint-manager.js
@@ -44,13 +44,16 @@ export class CuePointManager {
   }
 
   getCuePointById(id: string): ?TextTrackCue {
+    let cuePointById = null;
     const metadataTracks = this._getMetadataTracks();
-    metadataTracks.forEach(track => {
+    metadataTracks.every(track => {
       const cuePoint = track.cues.getCueById(id);
       if (cuePoint) {
-        return cuePoint;
+        cuePointById = cuePoint;
       }
+      return !cuePointById;
     });
+    return cuePointById;
   }
 
   removeCuePoint(cuePoint: TextTrackCue) {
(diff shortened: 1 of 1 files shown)


src/common/cuepoint/cuepoint-manager.js
```diff
@@ -44,13 +44,16 @@ export class CuePointManager {
...
-    metadataTracks.forEach(track => {
+    metadataTracks.every(track => {
       const cuePoint = track.cues.getCueById(id);
       if (cuePoint) {
-        return cuePoint;
+        cuePointById = cuePoint;
       }
+      return !cuePoint;
     });
+    return cuePointById;
```
RoyBregman, line 56:
```getCuePointById(id: string): ?TextTrackCue {
    const metadataTracks = this._getMetadataTracks();
    return metadataTracks.find(track => {
      return track.cues.getCueById(id);
    })
```

RoyBregman: We think this is clearer :)

RoyBregman (dismissed):

src/common/cuepoint/cuepoint-manager.js
```diff
@@ -44,13 +44,16 @@ export class CuePointManager {
   }
 
   getCuePointById(id: string): ?TextTrackCue {
+    let cuePointById = null;
     const metadataTracks = this._getMetadataTracks();
-    metadataTracks.forEach(track => {
+    metadataTracks.every(track => {
```
semarche-kaltura, line 49: I believe suggestion from Roy clearer or at  least you can use `metadataTracks.some` so on line 54 you can return `cuePoint` without `!`

src/common/cuepoint/cuepoint-manager.js
yairans: `metadataTracks.find` will return the track not the cue

src/common/cuepoint/cuepoint-manager.js
yairans: every instead of forEach it's just for breaking the loop once cue point is found, not related to the fix. I prefer it over some since im not looking for a track but over all of them

src/common/cuepoint/cuepoint-manager.js
RoyBregman: you are right

RoyBregman (approved):

semarche-kaltura (approved):

PR #524, opened 2022-02-08
fix(FEC-9716): An array doesn't merged into plugin config

SivanA-Kaltura:

If an array in plugin config contains an object item, call removeUnevaluatedExpression on it instead of filtering it out

Resolves FEC-9716

### CheckLists

- [ ] Docs have been updated

### DIFF
--- a/src/common/plugins/plugins-config.js
+++ b/src/common/plugins/plugins-config.js
@@ -30,7 +30,18 @@ const removeUnevaluatedExpression = (obj: Object = {}): Object =>
     if (Utils.Object.isObject(value) && typeof value !== 'function' && !Utils.Object.isClassInstance(value)) {
       product[key] = removeUnevaluatedExpression(value);
     } else if (Array.isArray(value)) {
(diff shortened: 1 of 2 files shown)


src/common/plugins/plugins-config.js
```diff
@@ -30,7 +30,18 @@ const removeUnevaluatedExpression = (obj: Object = {}): Object =>
     if (Utils.Object.isObject(value) && typeof value !== 'function' && !Utils.Object.isClassInstance(value)) {
       product[key] = removeUnevaluatedExpression(value);
     } else if (Array.isArray(value)) {
-      product[key] = value.filter(index => isValueEvaluated(index));
+      product[key] = value
+        .map(item => {
+          if (Utils.Object.isObject(item) && typeof item !== 'function' && !Utils.Object.isClassInstance(item)) {
```
yairans, line 35: Move this condition to const and reuse also above.
Also, can you move all this array handling to a function?

yairans (changes requested):

yairans (approved):

PR #576, opened 2022-09-22
fix(FEC-12347): sanitize before Json parse

lianbenjamin:
**the issue:**
in some cases we might get "invalid" data from BE, such as: entryName: 'invalid \t name'. at some point we are evaluating the data and executing Json.parse, which then fails; such failure is causing the player to not send analytics.

add sanitize function before calling Json.parse to omit invalid json chars.

Solves FEC-12347

### CheckLists

- [ ] Docs have been updated


giladna (approved):

test/src/common/plugin/plugins-config.spec.js
```diff
@@ -132,6 +132,14 @@ describe('evaluatePluginsConfig', () => {
     pluginsConfig.kava.should.have.property('playerVersion');
     pluginsConfig.kava.playerVersion.should.equal(__VERSION__);
   });
+
+  it('should sanitize the entryName', () => {
```
giladna, line 136: @lianbenjamin  what does this method is doing?

src/common/plugins/plugins-config.js
```diff
@@ -141,6 +143,30 @@ function getEncodedReferrer(): string {
...
 }
 
+/**
+ * @private
+ * @param {string} text - the string to sanitize
+ * @returns {string} - the sanitized string
+ * @private
+ */
+function _sanitize(text: string): string {
+  if (!text) return;
```
giladna, line 153: @lianbenjamin  this line fail in flow 
https://app.travis-ci.com/github/kaltura/kaltura-player-js/jobs/583634774

src/common/plugins/plugins-config.js
```diff
@@ -149,6 +175,7 @@ function getEncodedReferrer(): string {
  */
 function _formatConfigString(config: string): Object {
   let configObj;
+  config = _sanitize(config);
```
giladna, line 178: @lianbenjamin seems need to check here if config is not empty then you do not need the validation there 
and avoid parsing empty stuff

giladna (changes requested):

test/src/common/plugin/plugins-config.spec.js
lianbenjamin: it's a test- I pass an invalid payload and make sure it's sanitized and exist on the final object.

src/common/plugins/plugins-config.js
lianbenjamin: I check inside _sanitize method

giladna (approved):

PR #609, opened 2023-03-02
feat(FEC-10883): web media capabilities

lianbenjamin:

Expose new async API which will get the media capabilities from the navigator.
This API will return a response which indicates whether HEVC codec is supported and whether and which DRMs are supported in the browser.

Solves FEC-10883

### CheckLists

- [ ] Docs have been updated


src/kaltura-player.js
```diff
@@ -1191,6 +1191,124 @@ class KalturaPlayer extends FakeEventTarget {
...
+        videoCapabilities: [
+          {
+            contentType: 'video/mp4;codecs="avc1.42E01E"'
+          }
+        ]
+      }
+    ];
+
+    const keySystemsMap = new Map([
+      ['widevine', 'com.widevine.alpha'],
```
SivanA-Kaltura, line 1252: the list should be in consts.
and in a separate file, i think.
we shouldn't need to change kaltura-player.js every time we add a new type, only when we change the logic of the function.

SivanA-Kaltura (changes requested):

src/kaltura-player.js
```diff
@@ -1191,6 +1192,17 @@ class KalturaPlayer extends FakeEventTarget {
...
+
+  /**
+   * get the media capabilities
+   * @function getMediaCapabilities
+   * @param {HEVCConfigObject} hevcConfig - The HEVC configuration to check (optional).
+   * @returns {Promise<MediaCapabilitiesObject>} - The media capabilities object.
+   * @public
+   */
+  async getMediaCapabilities(hevcConfig?: HEVCConfigObject): Promise<MediaCapabilitiesObject> {
+    return await getMediaCapabilities(hevcConfig);
```
SivanA-Kaltura, line 1204: you don't need the await if you're just returning the promise

SivanA-Kaltura (changes requested):

docs/media-capabilities.md
```diff
@@ -0,0 +1,51 @@
...
+- `width` (number) - Optional width of the video. default is `1920`.
+- `height` (number) - Optional height of the video. default is `1080`.
+- `bitrate` (number) - Optional number of bits used to encode a second of video. default is `1200000`.
+- `framerate` (number) - Optional number of frames used in one second. default is `30`.
+
+The Media Capabilities API is retuning the following object structure:
+
+- `isHEVCSupported` (boolean | string) - Specifies whether HEVC is supported by the browser, or "maybe" if there was a problem with getting the information
+- `isPowerEfficient` (boolean | string) - Specifies power efficiency, or "maybe" if there was a problem with getting the information
+- `isDRMSupported` (boolean | string) - Specifies whether DRM is supported by the browser, or "maybe" if there was a problem with getting the information
```
semarche-kaltura, line 24: I would recommend use enums:
-1 = maybe
0 = no
1 =  true
or just use a string:
maybe = 'maybe'
no = 'no'
yes = 'yes'

better use one type instead of 2 types, like string or boolean

src/common/utils/media-capabilities.js
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * checks whether the browser supports HEVC codec or not
+ * @function _checkHEVCSupported
+ * @param {HEVCConfigObject} hevcConfig - The HEVC configuration.
+ * @returns {Promise<HEVCSupportedObject>} - The HEVC supported object.
+ * @private
+ */
+async function _checkHEVCSupported(hevcConfig?: HEVCConfigObject): Promise<HEVCSupportedObject> {
+  let hevcSupportedRes: HEVCSupportedObject = {
```
semarche-kaltura, line 100: actually `const` can be use here.
the object doesn't got re-defined, only key values

JonathanTGold (changes requested):

src/common/utils/media-capabilities.js
```diff
@@ -0,0 +1,143 @@
...
+    .decodingInfo(configHvc)
+    .then(result => {
+      hevcSupportedRes.isHEVCSupported = result.supported ? SupportedOptions.SUPPORTED : SupportedOptions.NOT_SUPPORTED;
+      hevcSupportedRes.isPowerEfficient = result.powerEfficient ? SupportedOptions.SUPPORTED : SupportedOptions.NOT_SUPPORTED;
+    })
+    .catch(() => {
+      hevcSupportedRes.isHEVCSupported = SupportedOptions.NOT_SUPPORTED;
+      hevcSupportedRes.isPowerEfficient = SupportedOptions.NOT_SUPPORTED;
+    });
+  return hevcSupportedRes;
```
JonathanTGold, line 140: `hevcSupportedRes` is sent before is it set at line 133/134

JonathanTGold (dismissed):

src/common/utils/media-capabilities.js
```diff
@@ -0,0 +1,143 @@
...
+      ]
+    }
+  ];
+
+  const keySystemsMap = new Map(DRM_SCHEME_LIST);
+
+  for (let [drm, keySys] of keySystemsMap) {
+    await navigator
+      .requestMediaKeySystemAccess(keySys, keySysConfig)
+      .then(() => {
```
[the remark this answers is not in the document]
lianbenjamin: @JonathanTGold  the `await` is for the entire result. the `then` is for the requestMediaKeySystemAccess API.. how are you suggesting it to look like?

src/common/utils/media-capabilities.js
lianbenjamin: @JonathanTGold it is not.. I just double checked. also, if that was the case, it would always return MAYBE, but it is returning the right value from my testing- SUPPORTED.

src/common/utils/media-capabilities.js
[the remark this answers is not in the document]
semarche-kaltura: @lianbenjamin, in theory it should work
try {
 await navigator.requestMediaKeySystemAccess(keySys, keySysConfig);
 drmSupportedRes.supportedDRMs.push(drm);
} catch(ex) {
_logger.debug(keySys + ' not supported (' + ex.name + ': ' + ex.message + ').');
}

semarche-kaltura (approved):

JonathanTGold (approved):

PR #627, opened 2023-05-31
fix(Sup:37514):Player V7 - Playlist side panel does not open for Live entries

inbalvasserman:
issue:
recordings for live in playlist take the live entry and not the recording entry

solution:
change merge deep to prefer to take the entry id returns from the server instead of the entry id saved in the playlist


SivanA-Kaltura:
resolved with a separate fix

PR #672, opened 2023-12-03
fix(FEC-13513): save playback speed to session storage

lianbenjamin:

save playbackRate in session storage instead of local storage.

Solves FEC-13513

### CheckLists

- [ ] Docs have been updated


src/common/utils/setup-helpers.js
```diff
@@ -123,8 +124,13 @@ function createKalturaPlayerContainer(targetId: string): string {
...
  */
 function setStorageConfig(options: KPOptionsObject): void {
-  if (!options.disableUserCache && StorageManager.isLocalStorageAvailable() && StorageManager.hasStorage()) {
-    Utils.Object.mergeDeep(options, StorageManager.getStorageConfig());
+  if (!options.disableUserCache) {
+    if (LocalStorageManager.isLocalStorageAvailable() && LocalStorageManager.hasStorage()) {
+      Utils.Object.mergeDeep(options, LocalStorageManager.getStorageConfig());
+    }
+    if (SessionStorageManager.isSessionStorageAvailable() && SessionStorageManager.hasStorage()) {
+      Utils.Object.mergeDeep(options, SessionStorageManager.getStorageConfig());
```
SivanA-Kaltura, line 132: I think setStorageConfig should also be by type. otherwise you're setting the same config into all storage managers, while currently its only relevant to local storage manager

src/common/storage/storage-wrapper.js
```diff
@@ -59,15 +63,16 @@ export default class StorageWrapper {
   /**
    * Gets an item from the local storage.
    * @param {string} key - The item key.
+   * @param {Object} storage - The storage object to use.
    * @static
    * @private
    * @returns {any} - The item value.
    */
-  static getItem(key: string): any {
+  static getItem(key: string, storage: Object = localStorage): any {
```
SivanA-Kaltura, line 71: I think a better solution would be to wrap
StorageWrapper.setItem in a function on the manager side, if you want to avoid passing the 2nd parameter every time.
I don't think the wrapper should know what kind of storage objects are available to it. just their API.

src/common/storage/base-storage-manager.js
```diff
@@ -0,0 +1,61 @@
...
+  static _buildStorageConfig(values: Object): Object {
+    const storageConfig = Utils.Object.mergeDeep({}, values);
+    delete storageConfig.textStyle;
+    return {
+      playback: storageConfig
+    };
+  }
+
+  static getStorageObject(): Object {
+    throw Error('');
```
SivanA-Kaltura, line 59: you can use something like this
https://github.com/kaltura/kaltura-player-js/blob/master/src/common/plugins/base-plugin.js#L66

src/common/storage/local-storage-manager.js
```diff
@@ -0,0 +1,114 @@
...
+  static getStorageObject() {
+    return localStorage;
+  }
+
+  /**
+   * @static
+   * @private
+   * @returns {boolean} - Whether local storage is implemented in the current browser.
+   */
+  static isLocalStorageAvailable(): boolean {
```
SivanA-Kaltura, line 27: i think we can just rename it to isStorageAvailable and put it on the interface, no ?
if we keep adding managers and every one of them will have isXStorageAvailable it will be redundant because we know what manager we are invoking this on

SivanA-Kaltura (changes requested):

src/common/storage/local-storage-manager.js
lianbenjamin: basically we will not have more storage managers that use this API because there are only 2- the session storage and local storage. other storages are being used in different ways or using different APIs. however, i do agree that it does make sense to move it to the base as well since both of them doing the same. will do it.

src/common/storage/storage-wrapper.js
lianbenjamin: hmm not sure i understand how the wrapper function on the manager side should look like and how it will help avoiding passing the storage as a parameter. either way eventually if the setItem (or getItem) is generic, then it has to receive the storage as well, no?
i thought it makes sense to have localStorage as the default, which means i don't need to change it on local storage manager side, but only on session storage manager side.

src/common/utils/setup-helpers.js
lianbenjamin: why it is setting the same config into all storage managers?
we first get the storage from local storage and merge it into options obj, then get the storage from session storage and merge it into options.

SivanA-Kaltura: yeah, you're right

src/common/storage/session-storage-manager.js
```diff
@@ -0,0 +1,31 @@
...
+import {BaseStorageManager} from './base-storage-manager';
+
+export default class SessionStorageManager extends BaseStorageManager {
+  static StorageKeys: {[key: string]: string} = {
+    PLAYBACK_RATE: 'playbackRate'
+  };
+  static _logger: any = getLogger('SessionStorageManager');
+
+  static getStorageObject() {
+    return sessionStorage;
```
MosheMaorKaltura, line 12: where this value is coming from?

src/common/utils/setup-helpers.js
```diff
@@ -135,8 +141,11 @@ function setStorageConfig(options: KPOptionsObject): void {
  * @returns {void}
  */
 function applyStorageSupport(player: KalturaPlayer): void {
-  if (StorageManager.isLocalStorageAvailable()) {
-    StorageManager.attach(player);
+  if (LocalStorageManager.isStorageAvailable()) {
```
MosheMaorKaltura, line 144: why does at this level, we need to know about storage types? why can we call - baseStorageManager::attachAll()
And it will check all available children and will attach each?

MosheMaorKaltura (approved):

SivanA-Kaltura (approved):

PR #731, opened 2024-01-31
chore: export storage managers from player

SivanA-Kaltura:

Expose managers and trigger type generation

Required for FEC-13504


JonathanTGold (changes requested):
We have a dedicated file for global types declarations
https://github.com/kaltura/kaltura-player-js/blob/master/src/types/global/globals.d.ts

JonathanTGold (commented):
what exactly the purpose os exporting these managers? 
I think the  browser storage should be managed only in  kaltura level

SivanA-Kaltura:
@JonathanTGold For [FEC-13504](https://kaltura.atlassian.net/browse/FEC-13504), I need to add code in a plugin to save a value in session config, and i want to do it through the session storage manager

JonathanTGold (approved):

PR #739, opened 2024-02-14
fix(SUP-41054): Playlist Chapter display issue with Live to VOD

Tzipi-kaltura:

**Issue:**
Chapters added to live entry (that currently not broadcasting) are not displayed when the entry is playing from playlist.

**Root cause:**
The sources.id on the player contains the live entryId and not the vod entryId so cuepoints request is done with the live entryId.
In the beginning the source.id = live entryId. as a result, in the merge (in setMedia) between sources from getPlaybackContext and localPlayer, the live entryId stay on the player and not replaced with the vod.

**Fix:**
If this is live entry but currently not broadcasting and vod is going to play, change the source.id to the vod id.

#### Resolves [SUP-41054](https://kaltura.atlassian.net/browse/SUP-41054)

[SUP-41054]: https://kaltura.atlassian.net/browse/SUP-41054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -182,7 +182,7 @@ export class KalturaPlayer extends FakeEventTarget {
     const playerConfig = Utils.Object.copyDeep(mediaConfig);
     //merge the current sources from player to keep the sources passed from constructor earlier
     const sources = Utils.Object.mergeDeep({}, playerConfig.sources, this._localPlayer.sources);
-    if (playerConfig.sources.type === MediaType.VOD && this._localPlayer.sources.mediaEntryType === MediaType.LIVE) {
+    if (playerConfig.sources.type === MediaType.VOD && this._localPlayer.sources.mediaEntryType?.toString() === MediaType.LIVE) {
       sources.id = playerConfig.sources.id;
     }
     delete playerConfig.sources;
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -182,7 +182,7 @@ export class KalturaPlayer extends FakeEventTarget {
     const playerConfig = Utils.Object.copyDeep(mediaConfig);
     //merge the current sources from player to keep the sources passed from constructor earlier
     const sources = Utils.Object.mergeDeep({}, playerConfig.sources, this._localPlayer.sources);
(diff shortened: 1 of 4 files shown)


src/kaltura-player.ts
```diff
@@ -182,6 +182,9 @@ export class KalturaPlayer extends FakeEventTarget {
     const playerConfig = Utils.Object.copyDeep(mediaConfig);
     //merge the current sources from player to keep the sources passed from constructor earlier
     const sources = Utils.Object.mergeDeep({}, playerConfig.sources, this._localPlayer.sources);
+    if (playerConfig.sources.type === 'Vod' && this._localPlayer.sources.mediaEntryType === 'Live') {
```
lianbenjamin, line 185: is there an enum to use for `Vod` and `Live` instead of hard-coded strings?
**Also**, can you please verify that this is not messing up https://github.com/kaltura/kaltura-player-js/pull/651 - when you switch to the live/vod entry in playlist, verify that the payload of `PLAYLIST_ITEM_CHANGED` contains `entryId` field with the "original" entry id (the live one).

lianbenjamin (changes requested):

src/kaltura-player.ts
Tzipi-kaltura: There is no enum, it's define as string.
Verified, it doesn't affect your patch, still send the live id.

lianbenjamin: there's `MediaType` enum. see https://github.com/kaltura/playkit-js/blob/074433785174ea69440693b5d59e3fa11268cf6b/src/enums/media-type.ts#L1
i think it is better to use the enum, you can do something like:
`playerConfig.sources.type === this.MediaType.VOD......` since there's a getter for `MediaType`.

Tzipi-kaltura: done

semarche-kaltura (dismissed):

MosheMaorKaltura (approved):

MosheMaorKaltura (approved):

lianbenjamin (approved):

PR #747, opened 2024-03-14
feat(FEC-13492): add support for prevent seek

lianbenjamin:

add support for prevent seek feature.

add listener to `prevent_seek` event and expose a getter so other entities can use it.

### DIFF
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -74,7 +74,8 @@ import {
   SourcesConfig,
   KPEventTypes,
   HEVCConfigObject,
-  MediaCapabilitiesObject
+  MediaCapabilitiesObject,
+  PreventSeekOptions
 } from './types';
 
 export class KalturaPlayer extends FakeEventTarget {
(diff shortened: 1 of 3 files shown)


JonathanTGold (changes requested):
I don’t think the player should be aware of the plugin, the all config should be managed inside the plugin, the plugin has access to the player instance

lianbenjamin:
@JonathanTGold the plugin is managing the configuration. the player listens to the event fired from the plugin and exposes API so other entities, such as kaltura cue points and timeline, will be able to use the API instead of listening to the event from the plugin.

src/kaltura-player.ts
```diff
@@ -854,6 +859,10 @@ export class KalturaPlayer extends FakeEventTarget {
         this._reset = false;
       }
     });
+    this._eventManager.listen(this, 'prevent_seek', (e: FakeEvent) => {
```
SivanA-Kaltura, line 862: hmmm
should ideally be a const exported from the plugin, but then we'll have a circular dependency...

SivanA-Kaltura (changes requested):

PR #765, opened 2024-04-11
feat(FEC-13871): set the entry name to navigator media metadata

lianbenjamin:

when playing a media on a mobile device and exiting the browser- the request is for the mobile device to show the entry name of the media that is played and the thumbnail as the background, in the native device player.

- using `MediaMetadata` object to pass to the `navigator.mediaSession`
- configuring the navigator `mediaSession` in `setMedia` to cover playlist and setMedia use-cases
- in case there is not entry name or poster, resetting to empty string so it will not show the previews information

### DIFF
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -169,6 +169,7 @@ export class KalturaPlayer extends FakeEventTarget {
       this._appPluginConfig = mergedPluginsConfigAndFromApp[1];
       this.configure(getDefaultRedirectOptions({ sources: this.sources }, mediaConfig));
       this.setMedia(mediaConfig);
+      this._configureMediaMetadata(mediaConfig);
       return mediaConfig;
     } catch (e) {
       const category = getErrorCategory(e);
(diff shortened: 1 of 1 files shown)


SivanA-Kaltura (changes requested):

SivanA-Kaltura (dismissed):

src/kaltura-player.ts
```diff
@@ -823,6 +824,14 @@ export class KalturaPlayer extends FakeEventTarget {
     return this._localPlayer.Error;
   }
 
+  private _configureMediaMetadata(mediaConfig: ProviderMediaConfigObject): void {
+    // here we can provide information about the media, to a device that is playing it
+    // set the media metadata title to the name of the entry
+    if (navigator.mediaSession) {
```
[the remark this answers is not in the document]
lianbenjamin: @JonathanTGold the requirement at the moment is for mobile but is it not limited to it. could also be TVs, tablets, etc..

src/kaltura-player.ts
```diff
@@ -823,6 +824,14 @@ export class KalturaPlayer extends FakeEventTarget {
     return this._localPlayer.Error;
   }
 
+  private _configureMediaMetadata(mediaConfig: KPMediaConfig): void {
```
JonathanTGold, line 827: if it's relevant only for mobile maybe it better to call it something like updateSessionForMobile() it would be more descriptive (the fact that it is configured media metadata can already be seen in the function declaration)

lianbenjamin: @JonathanTGold the requirement at the moment is for mobile but it is not limited to it. could also be TVs, tablets, etc..

JonathanTGold: got it bot anyway this is too generic name (and confusing with our media meta data object) it would be better to find a more specific name

semarche-kaltura (approved):

PR #776, opened 2024-05-21
fix(FEC-13878): copy the hls.worker.js file

lianbenjamin:

**the issue:**
for source only entries which we don't know where are the keyframes (i.e. zoom integration), the playback get stuck after making seek several times to different positions in the playback.

**solution:**
after investigating the issue, seems to only be an hlsjs issue, where hls is reducing the max buffer length to a value that is lower than the fragment duration.
when setting hls `enableWorker` configuration to `true`, the issue cannot be reproduced anymore.
at some point we disabled the worker since it stopped working and generated an error to the console (see relevant PR [here](https://github.com/kaltura/playkit-js-hls/pull/212/files)); in the current solution we make the worker work again by resolving the original issue of hls worker & webpak.

**related PR:** https://github.com/kaltura/playkit-js-hls/pull/218

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "source-map-loader": "^4.0.1",
     "standard-version": "^9.5.0",
     "style-loader": "3.3.3",
-    "terser-webpack-plugin": "^5.3.9",
+    "terser-webpack-plugin": "^5.3.10",
     "typescript": "^5.2.2",
     "webpack": "^5.88.2",
     "webpack-cli": "^5.1.4",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,6 +3,7 @@ const path = require('path');
 const CopyPlugin = require('copy-webpack-plugin');
 const packageData = require('./package.json');
 const chalk = require('chalk');
+const TerserPlugin = require('terser-webpack-plugin');
 
 module.exports = (env, { mode }) => {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -91,7 +91,7 @@
     "source-map-loader": "^4.0.1",
     "standard-version": "^9.5.0",
     "style-loader": "3.3.3",
-    "terser-webpack-plugin": "^5.3.9",
+    "terser-webpack-plugin": "^5.3.10",
```
SivanA-Kaltura, line 94: do we need to upgrade ?

SivanA-Kaltura (changes requested):

package.json
lianbenjamin: it is not necessary

webpack.config.js
```diff
@@ -26,6 +27,10 @@ module.exports = (env, { mode }) => {
     target: 'web',
     entry: './src/index.ts',
     devtool: 'source-map',
+    optimization: {
+      minimize: true,
```
JonathanTGold, line 31: True is the default and doesn’t have to be specified

package.json
JonathanTGold: @SivanA-Kaltura why not to upgrade?

MosheMaorKaltura (approved):

PR #792, opened 2024-06-05
feat(FEC-13946): Player core - Use shaka preload mechanism during playlist playback

SivanA-Kaltura:

- Add setCachedUrls API, which calls playkit (which calls dash adapter)
- Use API to prepare the next / previous playlist entries for a quick switch
- Change the behavior of session id assignment, so that the url we pass the setCachedUrls will be the same url that is later passed to adapter load()

Related PRs:
https://github.com/kaltura/playkit-js/pull/780
https://github.com/kaltura/playkit-js-dash/pull/262

Resolves FEC-13946

### DIFF
--- a/src/common/playlist/playlist-manager.ts
+++ b/src/common/playlist/playlist-manager.ts
@@ -389,7 +389,7 @@ class PlaylistManager {
 
           let sessionId = this._player.sessionIdCache?.get(mediaConfig.sources.id);
           if (!sessionId) {
-            sessionId = SessionIdGenerator.get();
+            sessionId = SessionIdGenerator.next();
             this._player.sessionIdCache?.set(mediaConfig.sources.id, sessionId);
           }
 
@@ -435,7 +435,7 @@ class PlaylistManager {
 
       let sessionId = this._player.sessionIdCache?.get(mediaConfig.sources.id);
       if (!sessionId) {
-        sessionId = SessionIdGenerator.get();
+        sessionId = SessionIdGenerator.next();
         this._player.sessionIdCache?.set(mediaConfig.sources.id, sessionId);
       }
 
--- a/src/common/playlist/playlist-manager.ts
+++ b/src/common/playlist/playlist-manager.ts
@@ -11,6 +11,8 @@ import { Playlist } from './playlist';
 import { PlaylistItem } from './playlist-item';
 import { mergeProviderPluginsConfig } from '../utils/setup-helpers';
 import { PlaylistOptions, PlaylistCountdownOptions, KalturaPlayerConfig, PluginsConfig, KPPlaylistObject, PlaylistConfigObject } from '../../types';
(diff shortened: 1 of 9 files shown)


src/common/utils/session-id-cache.ts
```diff
@@ -0,0 +1,17 @@
+class SessionIdCache {
+  private cache = new Map();
+
+  public set(key: string, value: string) {
```
MosheMaorKaltura, line 4: : void

src/common/utils/session-id-cache.ts
```diff
@@ -0,0 +1,17 @@
...
+    this.cache.set(key, value);
+  }
+
+  public get(key: string): string {
+    return this.cache.get(key);
+  }
+
+  public clear() {
```
MosheMaorKaltura, line 12: : void

src/common/utils/session-id-generator.ts
```diff
@@ -0,0 +1,28 @@
+import { Utils } from '@playkit-js/playkit-js';
+
+class SessionIdGenerator {
+  private static next: string = '';
+
+  private static init() {
+    SessionIdGenerator.next = `${Utils.Generator.guid()}:${Utils.Generator.guid()}`;
+  }
+
+  public static get() {
```
MosheMaorKaltura, line 10: this function is a getter but it creating a side effect of both init and setting next item.
Consider rename

src/common/playlist/playlist-manager.ts
```diff
@@ -388,6 +424,32 @@ class PlaylistManager {
   public destroy(): void {
     this._eventManager.destroy();
   }
+
+  private prepareEntry(index: number): Promise<any> {
+    if (this._playlist.items[index].isPlayable()) return Promise.resolve({ sources: this._playlist.items[index].sources });
```
MosheMaorKaltura, line 429: are you using prettier?

src/common/playlist/playlist-manager.ts
```diff
@@ -365,6 +386,21 @@ class PlaylistManager {
...
           this._playlist.updateItemSources(index, mediaConfig.sources);
           this._playlist.updateItemPlugins(index, mediaConfig.plugins);
+
+          let sessionId = this._player.sessionIdCache?.get(mediaConfig.sources.id);
+          if (!sessionId) {
+            sessionId = SessionIdGenerator.get();
+            this._player.sessionIdCache?.set(mediaConfig.sources.id, sessionId);
+          }
+
+          mediaConfig.sources.dash = mediaConfig.sources.dash.map((source) => {
```
MosheMaorKaltura, line 396: Is this only relevant for dash, seems that for now its usable for Dash but future wise its for all

src/common/playlist/playlist-manager.ts
SivanA-Kaltura: lint and prettier both, yes. they sometimes have conflicts. why, do you see an error here ?

PR #803, opened 2024-06-25
fix(SUP-43445): Chapter markers does not appear in player scrubber

Tzipi-kaltura:

**Issue:**
Chapter markers doesn't appear on the seekbar on live entry (that currently not broadcasting)

**Root cause:**
The marker displaying depend on duration promise, if duration from engine and player is not the same, the marker won't be displayed. 
souces.duration on the player contains the live duration and not the vod duration so in case the vod duration different from the live duration comparing the engine and source.duartion always returns false.

**Fix:**
If this is live entry but currently not broadcasting and vod is going to play, change the source.duration to the vod duration.

#### Resolves [SUP-43445](https://kaltura.atlassian.net/browse/SUP-43445)

[SUP-43445]: https://kaltura.atlassian.net/browse/SUP-43445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -186,6 +186,7 @@ export class KalturaPlayer extends FakeEventTarget {
     const sources = Utils.Object.mergeDeep({}, playerConfig.sources, this._localPlayer.sources);
     if (playerConfig.sources.type === MediaType.VOD && this._localPlayer.sources.mediaEntryType?.toString() === MediaType.LIVE) {
       sources.id = playerConfig.sources.id;
+      sources.duration = playerConfig.sources.duration;
     }
     delete playerConfig.sources;
     Utils.Object.mergeDeep(playerConfig.session, this._localPlayer.config.session);


src/kaltura-player.ts
```diff
@@ -186,6 +186,7 @@ export class KalturaPlayer extends FakeEventTarget {
     const sources = Utils.Object.mergeDeep({}, playerConfig.sources, this._localPlayer.sources);
     if (playerConfig.sources.type === MediaType.VOD && this._localPlayer.sources.mediaEntryType?.toString() === MediaType.LIVE) {
       sources.id = playerConfig.sources.id;
+      sources.duration = playerConfig.sources.duration;
```
SivanA-Kaltura, line 189: [live streams can have infinite duration](https://github.com/kaltura/playkit-js/blob/1463a414726752aa3b97d54b6d89b49b07d3ea90/src/engines/html5/media-source/adapters/native-adapter.ts#L1265)
why do you need to set the duration ? and why does it need to be on the source when the player has an api to retrieve duration ?
also live duration can change during runtime...

SivanA-Kaltura (changes requested):

src/kaltura-player.ts
Tzipi-kaltura: it's the same issue like - https://github.com/kaltura/kaltura-player-js/pull/739
we are talking about entries that were live in the past and now it's vod (the condition check this) but the live entryId in use so the sources.duration have the **live** duration but the compare in the timeline will be with the vod duration so need from the beginning use the vod duration

SivanA-Kaltura: got it
since this is the second time this code is updated - do you think there are any additional fields that are missing here ?

SivanA-Kaltura (changes requested):

src/kaltura-player.ts
Tzipi-kaltura: From my point of view there is nothing obvious, don't want to change if we are not sure.

SivanA-Kaltura (approved):

PR #818, opened 2024-07-16
feat(FEC-14049): add referrer to getPlaybackContext

MosheMaorKaltura:

Please add a detailed description of the change, whether it's an enhancement or a bugfix.

**Issue:**

**Fix:**

### DIFF
--- a/src/common/utils/kaltura-params.ts
+++ b/src/common/utils/kaltura-params.ts
@@ -127,8 +127,7 @@ function getReferrer(): string {
 
 function isInUnfriendlyIframe(): boolean {
   try {
-    window.parent.document.URL;
-    return false;
+    return (window.parent.document.URL !== window.parent.document.URL);;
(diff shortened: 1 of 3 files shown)


src/common/utils/kaltura-params.ts
```diff
@@ -125,6 +125,15 @@ function getReferrer(): string {
   return referrer;
 }
 
+function isInUnfriendlyIframe() {
+  try {
+    window.parent.document.URL;
```
SivanA-Kaltura, line 130: i dont really like a line that (on first glance) does nothing
how about this ?

try {
    return window.self.location.href !== window.top.location.href;
} catch() {
    return true;
}

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

PR #830, opened 2024-07-29
fix(FEC-14071): thumbnail requests append height, width, ks too many times

lianbenjamin:

bugfix

**Issue:**
when building `poster` and `sprite` URLs, we are using the `poster` as the base url and then appending params to it (height, width, ks). But, when poster has already been manipulated and we try to build it again, we re-append the params.

**Fix:**
use `rawThumbnailUrl` as the base url to build the urls.

related PR: https://github.com/kaltura/playkit-js-providers/pull/243

### DIFF
--- a/src/common/thumbnail-manager.ts
+++ b/src/common/thumbnail-manager.ts
@@ -71,11 +71,11 @@ class ThumbnailManager {
 
   private _buildKalturaThumbnailConfig = (uiConfig: UiConfig, mediaConfig: KPMediaConfig): KPThumbnailConfig => {
     const seekbarConfig = Utils.Object.getPropertyPath(uiConfig, 'components.seekbar');
-    const posterUrl = mediaConfig.sources && mediaConfig.sources.poster;
+    const rawThumbnailUrl = mediaConfig.sources && mediaConfig.sources.rawThumbnailUrl;
     const isVod = mediaConfig.sources && mediaConfig.sources.type === MediaType.VOD;
     const ks = this._player.shouldAddKs(mediaConfig) ? mediaConfig.session?.ks : '';
     const thumbnailConfig = Utils.Object.mergeDeep(DefaultThumbnailConfig, seekbarConfig);
-    const thumbsSprite = isVod ? this._getThumbSlicesUrl(posterUrl, ks, thumbnailConfig) : '';
+    const thumbsSprite = isVod ? this._getThumbSlicesUrl(rawThumbnailUrl, ks, thumbnailConfig) : '';
     return {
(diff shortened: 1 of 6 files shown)


src/common/thumbnail-manager.ts
```diff
@@ -71,11 +71,11 @@ class ThumbnailManager {
 
   private _buildKalturaThumbnailConfig = (uiConfig: UiConfig, mediaConfig: KPMediaConfig): KPThumbnailConfig => {
     const seekbarConfig = Utils.Object.getPropertyPath(uiConfig, 'components.seekbar');
-    const posterUrl = mediaConfig.sources && mediaConfig.sources.poster;
+    const rawThumbnailUrl = mediaConfig.sources && mediaConfig.sources.rawThumbnailUrl;
```
MosheMaorKaltura, line 74: mediaConfig.sources?.rawThumbnailUrl

tests/e2e/ovp/poster.spec.js
```diff
@@ -44,7 +44,7 @@ describe('addKalturaPoster', () => {
   });
 
   it('should append ks to kaltura poster', () => {
-    const mediaSources = { poster: '/p/1091/thumbnail/entry_id/0_wifqaipd/2' };
+    const mediaSources = { poster: '/p/1091/thumbnail/entry_id/0_wifqaipd/2', rawThumbnailUrl: '/p/1091/thumbnail/entry_id/0_wifqaipd/2'};
```
SivanA-Kaltura, line 47: note the lint errors

src/ovp/poster.ts
```diff
@@ -22,7 +22,8 @@ function addKalturaPoster(
   const playerWidth = dimensions.width;
   const playerHeight = dimensions.height;
   if (typeof playerPoster === 'string' && THUMBNAIL_REGEX.test(playerPoster) && playerPoster === mediaConfigPoster) {
```
SivanA-Kaltura, line 24: should the if condition be on rawThumbnailUrl instead ?
especially since we're testing poster for being set, but then in line 25 we use rawThumbnailUrl (which we didn't test whether it's set or not)

SivanA-Kaltura (changes requested):

src/ovp/poster.ts
lianbenjamin: hmmm yeah, it's a good idea to check the `rawThumbnailUrl`- i will **add** a check for it (not as a replacement to the `playerPoster`).

SivanA-Kaltura (approved):

PR #842, opened 2024-08-28
fix(SUP-44085): Two different players on the same page

Tzipi-kaltura:

**Issue:**
when there 2 or more different players on one page, video player is not loaded fine

**Fix:**
add the audio preset on new global variable according to the player targetId

pr related - https://github.com/kaltura/playkit-js-audio-player/pull/32/files

#### Resolves [SUP-44085](https://kaltura.atlassian.net/browse/SUP-44085)

[SUP-44085]: https://kaltura.atlassian.net/browse/SUP-44085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/ui-wrapper.ts
+++ b/src/common/ui-wrapper.ts
@@ -26,8 +26,8 @@ class UIWrapper {
       }
     } else {
       this._uiManager = new UIManager(player, config);
-      if (config.customPreset || window.kalturaCustomPreset) {
-        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPreset);
+      if (config.customPreset || window.kalturaCustomPresetMap[config.targetId]) {
+        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]);
       } else {
         this._uiManager.buildDefaultUI();
       }
--- a/src/common/ui-wrapper.ts
+++ b/src/common/ui-wrapper.ts
@@ -26,7 +26,7 @@ class UIWrapper {
       }
     } else {
       this._uiManager = new UIManager(player, config);
-      if (config.customPreset || window.kalturaCustomPresetMap[config.targetId]) {
+      if (config.customPreset || (window.kalturaCustomPresetMap && window.kalturaCustomPresetMap[config.targetId])) {
         this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]);
       } else {
         this._uiManager.buildDefaultUI();
--- a/src/common/ui-wrapper.ts
+++ b/src/common/ui-wrapper.ts
@@ -27,7 +27,7 @@ class UIWrapper {
     } else {
       this._uiManager = new UIManager(player, config);
       if (config.customPreset || (window.kalturaCustomPresetMap && window.kalturaCustomPresetMap[config.targetId])) {
-        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]['audioPreset']);
+        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]['audioPlayer']);
       } else {
         this._uiManager.buildDefaultUI();
       }
--- a/src/common/ui-wrapper.ts
+++ b/src/common/ui-wrapper.ts
@@ -27,7 +27,7 @@ class UIWrapper {
     } else {
       this._uiManager = new UIManager(player, config);
       if (config.customPreset || (window.kalturaCustomPresetMap && window.kalturaCustomPresetMap[config.targetId])) {
-        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]);
+        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]['audioPreset']);
       } else {
         this._uiManager.buildDefaultUI();
       }
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -426,6 +426,10 @@ export class KalturaPlayer extends FakeEventTarget {
     return this._localPlayer.getActiveTracks();
   }
 
+  public changeChildQuality(track: any | string): void {
+    this._localPlayer.changeChildQuality(track);
+  }
+
   public selectTrack(track: Track): void {
     this._localPlayer.selectTrack(track);
   }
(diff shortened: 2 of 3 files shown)


src/kaltura-player.ts
```diff
@@ -830,6 +831,15 @@ export class KalturaPlayer extends FakeEventTarget {
     return this._localPlayer.Error;
   }
 
+  private _addAudioPreset(options: Partial<KalturaPlayerConfig>): void {
+    if (options.plugins?.audioPlayer) {
```
SivanA-Kaltura, line 835: this just means that audio player is registered, not that it's loaded.
the plugin is only loaded after it's configuration is parsed successfully. 
are you sure that you want to run this code at this point already ?

also, just as a protection for the future, a better structure would be something like
window.kalturaCustomPresetMap[player id][preset name] = [preset function]
which allows us to support multiple custom presets, if we ever need them. otherwise, if we have only one kalturaCustomPreset variable, and we ever need to add another one, or have the player select between them, it will require a refactor.
but let me know how complicated it would be to make this change now, since it's not directly related to this PR, it's just future proofing

SivanA-Kaltura (changes requested):

src/kaltura-player.ts
Tzipi-kaltura: I thought it's good enough here cause we just need to make sure audio is exist, actually until now we add it in the beginning , now we add it in case it's audio player and not a video. i didn't think we should handle case it's not loaded because there is other issue but if you think we should change it, where i can put it, what does mean: "configuration is parsed"?
for you second comment - done, i changed it.

src/kaltura-player.ts
```diff
@@ -832,15 +836,17 @@ export class KalturaPlayer extends FakeEventTarget {
...
-        window.kalturaCustomPresetMap[options.ui!.targetId] = {};
+    Object.keys(options.plugins).forEach((plugin) => {
+      if (window.kalturaCustomPreset[plugin]) {
+        if (!window.kalturaCustomPresetMap) {
+          window.kalturaCustomPresetMap = {};
+        }
+        if (!window.kalturaCustomPresetMap[options.ui!.targetId]) {
+          window.kalturaCustomPresetMap[options.ui!.targetId] = {};
+        }
+        window.kalturaCustomPresetMap[options.ui!.targetId][plugin] = window.kalturaCustomPreset[plugin];
```
SivanA-Kaltura: won't this put window.kalturaCustomPreset[plugin] into every player on the window, which is what you want to avoid doing ?

src/kaltura-player.ts
```diff
@@ -832,15 +836,17 @@ export class KalturaPlayer extends FakeEventTarget {
   }
 
   private _addAudioPreset(options: Partial<KalturaPlayerConfig>): void {
```
SivanA-Kaltura, line 834: rename to _addCustomPreset or Presets if this becomes a general use function

src/kaltura-player.ts
```diff
@@ -426,6 +426,10 @@ export class KalturaPlayer extends FakeEventTarget {
     return this._localPlayer.getActiveTracks();
   }
 
+  public changeChildQuality(track: any | string): void {
```
SivanA-Kaltura: what does the function do ? because it's not clear from the name

src/kaltura-player.ts
```diff
@@ -832,15 +836,17 @@ export class KalturaPlayer extends FakeEventTarget {
...
-        window.kalturaCustomPresetMap = {};
-      }
-      if (!window.kalturaCustomPresetMap[options.ui!.targetId]) {
-        window.kalturaCustomPresetMap[options.ui!.targetId] = {};
+    Object.keys(options.plugins).forEach((plugin) => {
+      if (window.kalturaCustomPreset[plugin]) {
+        if (!window.kalturaCustomPresetMap) {
+          window.kalturaCustomPresetMap = {};
+        }
+        if (!window.kalturaCustomPresetMap[options.ui!.targetId]) {
```
SivanA-Kaltura: you can shorten to something like

window.kalturaCustomPresetMap[this.config.targetId] = window.kalturaCustomPresetMap[this.config.targetId] || {};

src/common/ui-wrapper.ts
```diff
@@ -27,7 +27,7 @@ class UIWrapper {
     } else {
       this._uiManager = new UIManager(player, config);
       if (config.customPreset || (window.kalturaCustomPresetMap && window.kalturaCustomPresetMap[config.targetId])) {
-        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]['audioPreset']);
+        this._uiManager.buildCustomUI(config.customPreset || window.kalturaCustomPresetMap[config.targetId]['audioPlayer']);
```
SivanA-Kaltura, line 30: don't use a hardcoded string for the plugin name, take it from an object like you do in kaltura-player.ts

SivanA-Kaltura (changes requested):

src/kaltura-player.ts
Tzipi-kaltura: no, the idea is to add the preset in the audio plugin into "audioPlayer" key so we don't need to search for specific plugin, this is also for future cases that we will add custom preset on on another places.
It's Moshe change

src/common/ui-wrapper.ts
Tzipi-kaltura: but i need specific preset, here is the audio preset, how can i use it from object?

src/kaltura-player.ts
Tzipi-kaltura: mistake, removed.

src/kaltura-player.ts
Tzipi-kaltura: done

src/kaltura-player.ts
Tzipi-kaltura: done

src/kaltura-player.ts
Tzipi-kaltura: checked again, this is not correct, in case we still don't have this player on kalturaCustomPresetMap we will get an error on the assignment:
image
image
I need the complete check, i can short it to:
`if (window.kalturaCustomPreset?.[plugin]) {
    window.kalturaCustomPresetMap ??= {};
    window.kalturaCustomPresetMap[options.ui!.targetId] ??= {};
    window.kalturaCustomPresetMap[options.ui!.targetId][plugin] = window.kalturaCustomPreset[plugin];
}
`
but i'm not sure if it's a clear code.
so up to you

SivanA-Kaltura (approved):

PR #914, opened 2025-02-01
fix(FEC-14289): Player v7 | Player doesn't recover well after network disconnection

roeedean:
After player had network disconnection error and user clicked on 'Try Again', player should be **destroyed** and we should **setup** a new player before using **loadMedia**.

Related PR: https://github.com/kaltura/playkit-js-ui/pull/978

Resolves [FEC-14289](https://kaltura.atlassian.net/browse/FEC-14289)

[FEC-14289]: https://kaltura.atlassian.net/browse/FEC-14289

### DIFF
--- a/src/common/utils/setup-helpers.ts
+++ b/src/common/utils/setup-helpers.ts
@@ -108,15 +108,15 @@ function validateProviderConfig(options: KalturaPlayerConfig): void {
 /**
  * Creates the player container dom element.
  * @private
- * @param {string} targetId - The div id which the player will append to.
  * @returns {string} - The player container id.
+ * @param options
  */
-function createKalturaPlayerContainer(targetId: string): string {
+function createKalturaPlayerContainer(options: PartialKPOptionsObject): string {
   const el = document.createElement('div');
-  el.id = Utils.Generator.uniqueId(5);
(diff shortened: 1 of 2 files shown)


src/common/utils/setup-helpers.ts
```diff
@@ -108,15 +108,15 @@ function validateProviderConfig(options: KalturaPlayerConfig): void {
 /**
  * Creates the player container dom element.
  * @private
- * @param {string} targetId - The div id which the player will append to.
  * @returns {string} - The player container id.
+ * @param options
  */
-function createKalturaPlayerContainer(targetId: string): string {
+function createKalturaPlayerContainer(options: PartialKPOptionsObject): string {
```
SivanA-Kaltura, line 114: can you explain why this was needed ?

SivanA-Kaltura (changes requested):

src/common/utils/setup-helpers.ts
roeedean: before destroying the old player we save his config.
when we are doing player.setup again after destroying the old player, we need access to the "options.ui.targetId" from the config.
we are creating a new player with the same config as the old one, so we need is element id to be the same.
if we won't use the old element id "options.ui.targetId", it will create the new player with unmatched element id which will cause errors later in the process of the player.setup.

PR #1203, opened 2026-04-26
feat(FEC-14949): Add entryId configuration support to Logo, Watermark, Error Slate and Bumper plugins

Tzipi-kaltura:

Adding support for entryId configuration in logo, errorOverlay and watermark.

checking components configuration to see if entryId is configured.
use getMediaConfig to get the img url.
update the components config with the new url.

related pr - https://github.com/kaltura/playkit-js-ui/pull/1160

#### Resolves [FEC-14949](https://kaltura.atlassian.net/browse/FEC-14949)

[FEC-14949]: https://kaltura.atlassian.net/browse/FEC-14949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/kaltura-player.ts
+++ b/src/kaltura-player.ts
@@ -1349,32 +1349,31 @@ export class KalturaPlayer extends FakeEventTarget {
   }
 
   private async _configureUiComponentsUrls(): Promise<void> {
-    const components = [
-      { name: 'logo', configKey: 'img' },
-      { name: 'watermark', configKey: 'img' },
-      { name: 'errorOverlay', configKey: 'backgroundUrl' }
-    ];
+    //@ts-expect-error - entriesForUiComponents
+    const entriesForUiComponents = this.config.playback?.entriesForUiComponents;
+    const data: Record<string, string> = {};
 
-    for (const { name, configKey } of components) {
-      const entryId = this.config.ui.components?.[name]?.entryId;
-      if (!entryId) continue;
-
-      try {
-        const mediaConfig = await this._provider.getMediaConfig({ entryId: entryId });
-        if (mediaConfig?.sources?.downloadUrl) {
-          this.configure({
-            ui: {
-              components: {
-                [name]: {
-                  [configKey]: mediaConfig.sources.downloadUrl
-                }
-              }
-            } as any
-          });
+    if (!entriesForUiComponents) {
+      return;
+    }
+    try {
+      const promises = Object.entries(entriesForUiComponents).map(async ([fieldName, entryId]) => {
+        try {
+          const mediaConfig = await this._provider.getMediaConfig({ entryId: entryId });
+          if (mediaConfig?.sources?.downloadUrl) {
+            data[fieldName] = mediaConfig.sources.downloadUrl;
+          }
+        } catch (error) {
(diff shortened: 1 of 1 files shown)


src/kaltura-player.ts
```diff
@@ -1346,4 +1347,34 @@ export class KalturaPlayer extends FakeEventTarget {
       } as any);
     }
   }
+
+  private async _configureUiComponentsUrls(): Promise<void> {
+    const components = [
```
SivanA-Kaltura: i think that you should put it in some central place in playkit-ui (redux state / ui config / event) and have ui take the values from there and put them into components, rather than having player be aware of what kind of components ui has.

SivanA-Kaltura (changes requested):

src/kaltura-player.ts
```diff
@@ -1346,4 +1347,34 @@ export class KalturaPlayer extends FakeEventTarget {
...
+          }
+        } catch (error) {
+          KalturaPlayer._logger.warn(`Cannot resolve entryId ${entryId} for UI component ${fieldName}`);
+        }
+      });
+      await Promise.all(promises);
+
+      if (Object.keys(data).length > 0) {
+        //@ts-expect-error - COMPONENTS_URLS_LOADED event
+        this.dispatchEvent(new FakeEvent(CoreEventType.COMPONENTS_URLS_LOADED, { data }));
```
SivanA-Kaltura: Call the event something like COMPONENT_DATA_UPDATED, i think, to be more generic.
Maybe we will reuse this for something that isn't urls

src/kaltura-player.ts
```diff
@@ -1346,4 +1347,34 @@ export class KalturaPlayer extends FakeEventTarget {
       } as any);
     }
   }
+
+  private async _configureUiComponentsUrls(): Promise<void> {
+    //@ts-expect-error - entriesForUiComponents
+    const entriesForUiComponents = this.config.playback?.entriesForUiComponents;
```
SivanA-Kaltura: actually, let's be organized and update 
kaltura-player-options.ts with a new top level config option called something like
config.uiComponentData
or
config.uiComponentDynamicData
or something similar

we don't want to update the playkit-js part of the config because playkit-js shouldn't "know" about ui

if you still get a type error after updating that file then you need to wait for the generated .d.ts file to be built, so add a TODO comment here like
"remove the ts expect error after type is updated"

src/kaltura-player.ts
SivanA-Kaltura:
```suggestion
        // this is an internal player event so it was not added to the type
        this.dispatchEvent(new FakeEvent(CoreEventType.COMPONENTS_URLS_LOADED, { data }));
```

SivanA-Kaltura: maybe we will add an "internal" or "other" event category eventually too, because we have events like this in other places

i think that the main criteria for whether we need to add an event to the type or not is whether it will be used by kava and/or whether a user will ever put an addEventListener for this event in their own code. it's not really relevant here, so it's not as important to add it to type.

SivanA-Kaltura (changes requested):

SivanA-Kaltura (approved):

roeedean (approved):

PR #1243, opened 2026-07-15
fix(player): guard against empty messages array in error-helper (SUP-52847)

inbalvasserman:
## Problem

Geo-blocked VOD content outside Israel (customer 2748741 / Reshet / 13tv.co.il) shows an infinite loader instead of the proper geo-block / access-control error screen.

## Root Cause

In `src/common/utils/error-helper.ts`, all 5 restriction checks used this pattern:

\`\`\`ts
error.data?.messages && error.data?.messages[0].code === 'X'
\`\`\`

When the Kaltura API returns a BLOCK action with `messages: []` (empty array), `error.data?.messages` evaluates to `[]` which is **truthy** in JavaScript. The chain then proceeds to `messages[0]` which is `undefined`, and `undefined.code` throws a **TypeError**.

This uncaught TypeError propagates out of `getErrorCategory()`, preventing the ERROR event from ever being dispatched. Result: `hasError` stays `false`, the loading spinner stays visible indefinitely — the error screen never renders.

## Solution

Replaced the unguarded array access with optional chaining in all 5 restriction checks:

\`\`\`ts
// Before
error.data?.messages && error.data?.messages[0].code === 'X'

// After
error.data?.messages?.[0]?.code === 'X'
\`\`\`

When `messages: []`, the checks now safely return `false` instead of throwing, allowing `isAccessControlRestrictedError` to match and the correct error screen to render.

## Test Coverage

- [x] Added regression test: `tests/e2e/common/utils/error-helper.spec.ts`
- [x] All existing tests pass
- [x] Manual verification: BEFORE build shows infinite loader, AFTER build shows error screen

## Related

- Jira: https://kaltura.atlassian.net/browse/SUP-52847

## Checklist

- [ ] Code reviewed
- [ ] No regressions introduced
- [ ] Test covers the failure case

### DIFF
--- a/src/common/utils/error-helper.ts
+++ b/src/common/utils/error-helper.ts
@@ -5,11 +5,11 @@ const isBlockAction = (error: Error): boolean => error.code === 2001;
 const isMediaNotReady = (error: Error): boolean => error.code === 2002;
 const isScheduledRestrictedCode = (error: Error): boolean => error.code === 2003;
 const isDeletedEntryCode = (error: Error): boolean => error.code === 2004;
-const isGeolocationRestricted = (error: Error): boolean => error.data?.messages && error.data?.messages[0].code === 'COUNTRY_RESTRICTED';
-const isSessionRestricted = (error: Error): boolean => error.data?.messages && error.data?.messages[0].code === 'SESSION_RESTRICTED';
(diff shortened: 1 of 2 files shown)


inbalvasserman:
Superseded by PR from branch fix/SUP-52847

